PR #24, opened 2016-02-17
Add postinstall

wpears:
Tiny change for automatic scaffolding of dist (postinstall runs automatically after install)


awolfe76:
:+1:

Issue #3, opened 2016-01-29
Implement basic file submission page


wpears:
Basic functionality is allowing the submission of a file and signaling to the user when it has been fully received by the server.

awolfe76:
#14 has a possible component breakdown for this page.

awolfe76:
#19 started basic framework for this.

awolfe76:
@wpears - I think this is what you are currently working on, right? And #28 might complete this?

wpears:
Yeah will be closed by #28

awolfe76:
#28 is merged, closing.

PR #28, opened 2016-02-29
Upload component

### LINKED ISSUE #21: Upgrade submission form
Initially, this will be a form that:
- Allows submission
- Tracks upload progress
- Indicates when successfully received

We'll need the [xhr progress event](https://developer.mozilla.org/en-US/docs/Web/API/ProgressEvent) to track the upload progress, which of course means we'll be using XHR over vanilla form submission.

There is the opportunity to allow for a chunked-upload, where we split the file and only stream over pieces at a time, allowing easy restart if an upload is interrupted, but this requires much more server-side sophistication which is better spent elsewhere (unless we find that interrupted uploads are 1. common and 2. slow enough to warrant this affordance).

wpears:
Basic, implementation of upload component, closes #21


wpears:
Also closes #3

package.json
```diff
@@ -59,5 +59,9 @@
       "<rootDir>/node_modules/react-addons-test-utils",
       "<rootDir>/node_modules/fbjs"
     ]
+  },
+  "dependencies": {
+    "react": "^0.14.7",
```
awolfe76, line 62: These being here is right, but should we leave them in `devDependencies` too? Lines 43 and 45.

src/js/makeSubmitContainer.jsx
```diff
@@ -0,0 +1,85 @@
...
+      var className = '';
+      var file = this.props.file;
+      var size = 0;
+      if(!file){
+        className = 'hidden';
+      }else{
+        size = file.size;
+      }
+      return (
+        <div id="uploadTracker" className={className}>
```
awolfe76: I like this with the size, but wonder if we should do a progress bar (or something)?
This isn't holding up a merge though, we can always change it later.

__tests__/submitContainer.js
```diff
@@ -1,15 +1,16 @@
...
 var TestUtils = require('react-addons-test-utils');
 
-var FileSubmissionForm = require('../src/js/file-submission-form.jsx');
+jest.autoMockOff();
+
+var makeSubmitContainer = require('../src/js/makeSubmitContainer.jsx');
+var SubmitContainer = makeSubmitContainer();
 
-describe('file-submission-form', function() {
+describe('submitcontainer', function() {
```
awolfe76: Can we add a few tests?
- file selection
- form submission

awolfe76:
Nice! Just a few things there @wpears.

src/js/makeSubmitContainer.jsx
wpears: Yeah, this will be solved with the PR that addresses https://github.com/cfpb/hmda-platform-ui/issues/34

awolfe76:
:+1:

PR #52, opened 2016-03-17
Homepage components

wpears:
Contains components for the homepage, including the container for the FIs, headers, the fiStatus components, links which enable easy routing and others.

Slipped some extra tests in for great justice


src/js/index.js
```diff
@@ -1,12 +1,31 @@
...
+var fiContainer = require('./fiContainer.jsx');
+//var makeSubmitForm = require('./makeSubmitForm.jsx');
 
+/*
 var SubmitForm = makeSubmitForm('/submit', function(e){
   console.log(e, 'file uploaded');
 });
+*/
 
 ReactDOM.render(
```
awolfe76, line 8: We won't need this later, or at least in production, do you have thoughts on handling rendering this for dev?

src/js/institutions.js
```diff
@@ -0,0 +1,140 @@
+module.exports = {
```
awolfe76: Wouldn't hold up this PR for it, but we should probably move this, and others we'll be creating, into a subdirectory.

src/js/index.js
wpears, line 8: Well actually, I think we will need it for production, just probably not in index.js. That's the submit component aye?
We'd just need to route to it and it probably/possibly won't be included here.

I'll nix it from index.js

awolfe76, line 8: Sorry, meant that on the userSelect stuff. We wouldn't have a dropdown to choose the user there.

src/js/fiContainer.jsx
```diff
@@ -0,0 +1,59 @@
...
+var divisionHeader = require('./divisionHeader.jsx');
+var fiStatus = require('./fiStatus.jsx');
+
+var fiContainer = React.createClass({
+
+  getInitialState: function(){
+    return this.updateDivisions(this.props.institutions);
+  },
+
+  updateDivisions: function(institutions){
```
awolfe76: Can we use a different name other than `updateDivisions`?

Seems like this is grouping each institution by the status, `updateInstitutionsByStatus` maybe? Or even `sortInstitutionsByStatus`?

If so, I think the `state.divisions` array below (~line 13) could be changed, and the `divisionHeader.jsx` filename and vars within.

src/js/fiContainer.jsx
```diff
@@ -0,0 +1,59 @@
...
+    return (
+      <div id="fiContainer">
+        {this.state.divisions.map(function(division, i){
+          var header = null;
+          if(division.institutions.length) header = React.createElement(divisionHeader, {text: division.text});
+          return (
+            <div key={i} className="divisionWrapper">
+              {header}
+              {division.institutions.map(function(institution, i){
+                return React.createElement(fiStatus, {key: i, institution: institution});
```
awolfe76: Is there a huge difference in using `React.createElement` to handle this vs something like
`<fiStatus key={i} institution={institution} />`?

src/js/fiStatus.jsx
```diff
@@ -0,0 +1,113 @@
...
+      num = timeSince/60000 >> 0
+    }
+
+    if(num === 1) unit = unit.slice(0, -1);
+
+    return filingString + num + unit + ' ago.'
+  },
+
+  resubmitComponent: function(){
+    return <div className="resubmitComponent">{React.createElement(cbLink, {text: 'Resubmit', callback: router.resubmit})} - Resubmitting will allow you to correct any errors or invalid data encountered. Each error report will be saved so you can track your progress after resubmission.</div>
```
awolfe76: Same as from `fiContainer.jsx` - could we do `<cbLink text='Resubmit' callback={router.resubmit} />`?

src/js/fiStatus.jsx
```diff
@@ -0,0 +1,113 @@
...
+        })}
+      </ul>
+    )
+  },
+
+  getStatusText: function(statusCode){
+    switch(statusCode){
+      case 0:
+        return (
+          <div className="fiStatusWrapper">
```
awolfe76: We could move the

```
<div className="fiStatusWrapper">
<div className="fiStatusText">
```

and 

```
</div>
</div>
```

to the `render` function to avoid the repetition in the case statements.

Maybe to handle

```
{this.resubmitComponent()}
{this.editReportsComponent()}
```

we could check the status in `render` too?

src/js/index.js
```diff
@@ -1,12 +1,25 @@
...
-var makeSubmitForm = require('./makeSubmitForm.jsx');
+var api = require('./api');
+var userSelect = require('./userSelect.jsx');
+var fiContainer = require('./fiContainer.jsx');
 
-var SubmitForm = makeSubmitForm('/submit', function(e){
-  console.log(e, 'file uploaded');
-});
 
 ReactDOM.render(
```
awolfe76, line 8: Sorry @wpears. That comment early was for this `userSelect`. We won't need it later, or in production, right?

src/js/cbLink.jsx
```diff
@@ -0,0 +1,12 @@
+var React = require('react');
```
awolfe76: Just curious, do we _need_ this component? Couldn't we just render links as normal with JSX from within the other components?

Doing it this way, should we use this component to render all links in the app?

src/js/fiStatus.jsx
```diff
@@ -0,0 +1,113 @@
...
+  resubmitComponent: function(){
+    return <div className="resubmitComponent">{React.createElement(cbLink, {text: 'Resubmit', callback: router.resubmit})} - Resubmitting will allow you to correct any errors or invalid data encountered. Each error report will be saved so you can track your progress after resubmission.</div>
+  },
+
+  editReportsComponent: function(){
+    return (
+      <ul className="editReportsWrapper">
+        {this.state.editReports.map(function(report, i){
+          return (
+            <li key={i} className="editReport">
```
awolfe76: Can you tell me why we need the `key` here and not in `resubmitComponent` above? Not sure I've wrapped my head around that yet.

awolfe76:
@wpears I think I'm done ... for now. :smiling_imp: 

Really cool stuff here. Nice work. :rocket:

src/js/fiContainer.jsx
wpears: I thought not, yet I was experiencing bugs when I had custom components declared with JSX (bizarrely, because it is certainly supposed). React.createElement removed the trouble so I stuck with that mode.

src/js/index.js
wpears, line 8: Ahhh righto, thought that might be the case. Yeah we won't. We won't need to replace it though until we get the auth component hammered out, so I say leave it in for now.

src/js/cbLink.jsx
wpears: The main purpose here was as a place for routing to hook into, or for common onClick code to be executed (like e.preventDefault()). Not strictly needed, but I think as we implement routing we'll want something like this.

src/js/fiStatus.jsx
wpears: React needs it when you are rendering dynamic children https://facebook.github.io/react/docs/multiple-components.html#dynamic-children

src/js/fiContainer.jsx
awolfe76: Doing it like this works `<FIStatus key={i} institution={institution} />` with a capital letter first. The lower case version doesn't throw an error :frowning: but doesn't render either.

Of course we'd have to update the require as well, `var FIStatus = require('./fiStatus.jsx');`

Unless you want to fix it now, I'm ok merging this the way it is and we can create an issue to come back and apply the JSX way. I'll handle that, will give me a chance to read through the PR in more detail too.

wpears: Hah! Fixing now.

awolfe76:
Nice ... merging.

PR #54, opened 2016-03-23
User, FI heading

### LINKED ISSUE #29: Begin creating header component
For now, the header component is pretty straightforward. On the home page we'll render something like:
`Welcome to {this.props.year} HMDA filing, {this.props.user}.`

header

While filing this content will change slightly to:
`{this.props.user} filing in {this.props.year} on behalf of {this.props.institution}`

header2

The `user` will come from our authentication flow - #7.
The `year` will be set for our first year, but later filers will be able to select the year to allow for resubmissions.
The `institution` on filing screens will be set when selected on the home page.

awolfe76:
Adds a heading to display the user, FI, and year for the filer.

The component is created within `index.js` in the callback for the user selection. I think that's ok, for now, because we don't have another way to get a user.

The year is also hard coded. For the first time we collect files this is also ok, only 1 year (2017) will be collected.

Closes #29 (see for more details)

### DIFF
--- a/__tests__/UserFiHeading.js
+++ b/__tests__/UserFiHeading.js
@@ -24,31 +24,66 @@ var institution = {
   ]
 }
 
-describe('FiStatus', function(){
+describe('UserFiHeading', function(){
+  describe('render without institution', function() {
 
-  var headingComponent = <UserFiHeading institution={{}} year='2017' user={user} />
+    var headingComponent = <UserFiHeading institution={{}} year='2017' user={user} />
 
-  var heading = TestUtils.renderIntoDocument(headingComponent);
-  var headingNode = ReactDOM.findDOMNode(heading);
+    var heading = TestUtils.renderIntoDocument(headingComponent);
+    var headingNode = ReactDOM.findDOMNode(heading);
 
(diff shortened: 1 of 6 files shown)


awolfe76:
Added a small one to watch the `__tests__` for changes too.

wpears:
I think this should be a child of FiContainer. This way you could hook into the state that it manages (specifically institutions) and also be rendered as part of the same call to ReactDOM.render.

Thoughts?

awolfe76:
I had that thought too, but it's also rendered outside of that container on the other "pages".

For example, after they select an FI they would see something like `[name] filing in [year] on behalf of [FI]`. In this case we don't need the entire list of institutions, just the one they've selected. And I don't think that FiContainer is used in that case since its just the list of all possible FI's and their current status.

Am I right about that?

wpears:
Hmm. That's true. Kinda shows a need for a page level container that... contains... everything. And that is the thing that will manage user state, etc. and pass them as props to both the FiContainer and this header.

Gonna ticket that, and merge this.

awolfe76:
:+1:

PR #56, opened 2016-03-24
Start adding CF

### LINKED ISSUE #23: Include Capital Framework
Add CF to the project.

awolfe76:
This is just the first step in this. We will be adding in other cf- modules soon, but wanted to at least get this up. This closes #23 ... but we'll have more additions to this which we can create issues for as needed.
- adds in less for building css (and remove node-sass)
- updates to build process to handle it
- adds in cf-core, cf-grid, cf-typography
- started app.less
- updates to components and tests to get it all going

The "home" page now looks like:
styles


wpears:
Great work on this Andrew; certainly a welcome addition. My one qualm is that I think we should keep the component classes, especially for how it interacts with testing parent components with lots of rendered children (can't check the number of say, <p>s rendered if it's rendered children have <p>s as well... though there could be solutions involving more complicated, depth-based queries which would work).

At this stage, I also think they're kinda nice as debugging handles from the console. Even if they never show up in `app.less`, I think they have merit.

awolfe76:
Makes sense, I'll add them back in. Do they need to match the component name exactly though, since they are hard coded in the tests anyway? Not sure if you are using them for anything else.

awolfe76:
In looking again, it looks like there are only 2 not there, `cbLink` and `divisionHeader`.

For `cbLink` I can add a class to those links to double check but is that component going away when we implement router? Looks like router uses a `<Link>` tag, so might be easier to move links back in line rather than using a component?

wpears:
Yeah, you're right, I am changing stuff whilst implementing the router.

Merging; if I want class stuff I'll add it in there.

PR #59, opened 2016-03-25
Routing

wpears:
Introduces react router to control history and move between app screens (try clicking the begin filing now link).

Also adds a page level component for managing state.


src/js/AppContainer.jsx
```diff
@@ -0,0 +1,45 @@
...
+
+  setUser: function(user){
+    this.setState({user: user});
+  },
+
+  render: function() {
+    return (
+      <div>
+        <UserSelect callback={this.selectCallback}/>
+        <UserFiHeading institution={{}} year="2017" user={this.state.user.name}/>
```
awolfe76: Everything works when clicking the "Begin filing now" link to see the upload form except that the heading, `UserFiHeading.jsx`, doesn't get updated. Looks like its because its always getting an empty object, `{}` for `institution`. 

Do we need to add a value to the `state` to hold the institution when one is selected?

Or does the `<Link>` in `FiStatus.jsx` (~line 66) need to be something like:
`<p><Link to="/upload/{this.props.institution.name}">Begin filing now</Link>.</p>`
And then the `<Route>` in `index.js` (~line 22) changed to:
`<Route path="/upload/:institutionName" component={Submit}/>`

src/js/AppContainer.jsx
```diff
@@ -0,0 +1,45 @@
...
+        {React.cloneElement(
```
awolfe76, line 45: I'm not quite following what the `React.cloneElement` is doing.

src/js/AppContainer.jsx
wpears: I like the second options, adding now.

src/js/AppContainer.jsx
wpears, line 45: Yeah sorry, I'll add in a comment.

Because this component is set up by a Route, there isn't a natural way to pass in props that get sent to dynamic children except by doing this. The children and cloned and get `institutions: this.state.user.institutions` merged into whatever props they've been given.

wpears, line 45: This also could have been done by having a wrapper around the components and keeping variables in its closure, but I think the cloneElement way is somewhat more idiomatic.

src/js/AppContainer.jsx
wpears: Acttttually, the FiStatusHeader is rendered in `AppContainer` unconditionally, which is rendered on every route, so the route would need to passed in at a higher level.

I like the idea of the route in the URL, and we could do that, but it would definitely need to be encoded with encodeURIComponent.. Thoughts?

awolfe76: Hmm, I think I see what you mean.

But the only time that we care to render the institution name is on any page but the home page. So, maybe its ok this way? We render the `UserFiHeading` no matter what, but it gets sent an institution if `this.props.params.institution` is set or it gets sent the default empty object, `{}` like it does now.

Or am I missing something else?

wpears: Actually this _is_ solvable with how the router works (I deleted my previous comment, but not quickly enough ;)), will push up soon (just working on some stuff related to the UserSelect).

src/js/AppContainer.jsx
```diff
@@ -41,6 +41,7 @@ var AppContainer = React.createClass({
       <div>
         {selectComponent}
         <UserFiHeading institution={{name: params.encodedInstitutionName}} year="2017" user={this.state.user.name}/>
+        /*clones any children added by the router in order to pass them props*/
```
awolfe76, line 44: Your comment needs to be `{/*clones any children added by the router in order to pass them props*/}` (wrapped in `{}`) so that it doesn't show up in the UI.

awolfe76:
After the comment fix you're good to merge @wpears!

src/js/AppContainer.jsx
awolfe76, line 44: Never mind, I see you fixed it in the next PR.

awolfe76:
Merging, the comment is fixed in the next PR.

PR #69, opened 2016-03-30
Cleanup

wpears:
Addresses the naming consistency / monolithic component / extraneous file issues in #64, #65, #66, and #68.

Depends on #59

Issue #67, opened 2016-03-29
Create Tracker or Progress component


awolfe76:
This is currently in [`makeSubmitForm.jsx`](https://github.com/cfpb/hmda-platform-ui/blob/master/src/js/makeSubmitForm.jsx#L54), `createClass` is used. We can break out this component and allow it to be used in a few more cases, basically where any `x of y has been checked`.

wpears:
Also, when this is stripped out, turn makeUploadForm into a component rather than a factory, as discussed in #70

awolfe76:
Closed by #71

PR #70, opened 2016-03-30
CSS

### LINKED ISSUE #63: CSS
Begin creating `.less` files for `custom.less`, containing non-component related extensions to CF and `[component-name].less` for each component as needed.

These new files will contain component specific styles. See #60.

awolfe76:
- adds hmda and components directory
  - hmda to hold CF overrides and styles based on CF modules
  - components for react component specific styles
- rename `makeUploadForm` to `uploadForm`

closes #63


wpears:
So I didn't switch `makeUploadForm` as part of #69 because it returns a factory rather than a component. It _could_ return a component... and maybe should.. but I think that having a factory with the component naming scheme is somewhat confusing

awolfe76:
Ahh, I see now. Switched back.

Any reason not to make it a component, maybe add that to #67?

wpears:
Roger, added to #67, which I'll take care of.

PR #71, opened 2016-04-01
Progress tracker

wpears:
Implements a progress tracking component and reworks the UploadForm to use it.


src/js/Progress.jsx
```diff
@@ -0,0 +1,43 @@
+var React = require('react');
+
+var Progress = React.createClass({
+  propTypes: {
+    progress: React.PropTypes.number,
```
awolfe76, line 5: This is interesting. I'll create a ticket to go back to other components and add this.

PR #72, opened 2016-04-01
Homepage link component

wpears:
Depends on #71 
Small Homepage link component, rendered as a part of the AppContainer


src/less/components/homeLink.less
```diff
@@ -0,0 +1,5 @@
+.HomeLink {
```
awolfe76: I think we should just remove this `.less` file for now. This should probably render in place acting almost like a bread crumb. See this example - http://beta.consumerfinance.gov/the-bureau/leadership-calendar/

home-link

We can add styling later ... to add the icon if needed.

src/less/app.less
```diff
@@ -14,6 +14,7 @@
 @import (less) "components/institutionStatus.less";
 @import (less) "components/resubmit.less";
 @import (less) "components/uploadForm.less";
+@import (less) "components/homeLink.less";
```
awolfe76: Not needed if we remove the `.less` file.

src/js/HomeLink.jsx
```diff
@@ -0,0 +1,10 @@
+var React = require('react');
+var Link = require('react-router').Link;
+
+var HomeLink = React.createClass({
+  render: function(){
+    return <Link className="HomeLink" to="/">Home</Link>
```
awolfe76, line 6: We can remove the `className="HomeLink"` here. We _may_ add another class at some point to add an icon.

wpears, line 6: I kinda like having the components name as the classname on the node it returns. Seems convenient for debugging and such, and any specific styling outside the general grid system, etc. Doesn't really cost us anything.

awolfe76, line 6: :+1: on the classname!

wpears, line 6: Ticketed as #74

awolfe76:
:100:

PR #76, opened 2016-04-05
Edits

### LINKED ISSUE #42: Component to render syntactical and validity edits
The links that cause these components to render is from #35.

syn-val

This table can be used for both syntactical and validity edits.
Initially show the rows with a column containing loan number and then number of errors. The number of errors will be a link to the more detailed information.

Data required:
- loan number
- number of errors
- edit id
- edit description
- field where the error occurred
- the incorrect data that was submitted
- _action to be taken (this may not be required, all errors in syntactical and validity require correction and for the file to be re-uploaded_

While there may be no link generated from #35, there are no errors present, this component should still be capable of showing a message. A filer could still navigate to `/edits/syntactical` or `/edits/validity` (see #37) so they should see a message saying something like `There are no [editType] errors. Please continue to the next step.`.

awolfe76:
- starts edit screens for syntax and validity edits
- adds CSS for them
- has hardcoded data for now (see #77)
- will also break out a few more components (see #78 and #79)
- for now the route is `/edits/` (may change in a future PR)
- results are currently all expanded (see #80)

Screenshot:

edits

Closes #42

### DIFF
--- a/src/js/EditsContainer.jsx
+++ b/src/js/EditsContainer.jsx
@@ -13,7 +13,7 @@ var EditsContainer = React.createClass({
     }
   },
 
-  componentDidMount: function() {
+  componentWillMount: function() {
     var _this = this;
     _this.setState (
       {
--- a/src/js/EditsContainer.jsx
+++ b/src/js/EditsContainer.jsx
@@ -1,15 +1,15 @@
 var React = require('react');
-var ErrorsSyntax = require('./ErrorsSyntax.jsx');
+var EditsSyntacticalValidity = require('./EditsSyntacticalValidity.jsx');
 
-var ErrorsContainer = React.createClass({
+var EditsContainer = React.createClass({
   getInitialState: function() {
     return {
-      errors: {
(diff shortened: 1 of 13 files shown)


awolfe76:
@wpears - should be ok to review now. I've created a few other issues to clean up some items in this PR.

src/js/EditsContainer.jsx
```diff
@@ -0,0 +1,128 @@
...
+    	"edits": {
+    		"syntactical": [],
+    		"validity": [],
+    		"quality": [],
+    		"macro": []
+    	}
+    }
+  },
+
+  componentDidMount: function() {
```
wpears: `componentDidMount` should be `componentWillMount`, otherwise you'll be waiting on the first render before setting the initial data

wpears:
Looks good, just the one comment on componentDidMount

awolfe76:
Good call @wpears. I'll merge once Travis gives the ok.

PR #85, opened 2016-04-08
Macro edits / rework EditDetail

### LINKED ISSUE #45: Component to render macro edits
While this component will act a lot like the one in #42 and #44 I think its different enough to justify being on its own.

Macro edits are _not_ based on row-by-row, but instead are based on the file as a whole and external resources such as the previous years filing.
### _With_ syntactical and validity errors

macro

If syntactical and validity errors are still present, a informational message should appear to the filer.

This table is different from others, it's no longer row-by-row, instead it is edit-by-edit. The component renders the edit ID, the edit description, and whether or not it has been verified. Click the edit ID expands  to show more detail. 

Data required:
- edit ID
- edit description
- verified or not
- details about the error
  - for example, for Q007 we would want to tell them that they had 252 / 551 loans that had an action code = 2. So that's 45.75% which is > than 15%.
### _Without_ syntactical and validity errors

Once the syntactical and validity edits have run without error, this component renders a bit differently. The informational message is removed and now a reason dropdown is rendered, along with a checkbox and verification message.

@debseidner @lesgou - are these "reason dropdowns" always the same?

_The checkbox and verification message should **not** show up until a reason has been selected._

macro-verify

Once a reason has been selected and the checkbox checked only the corresponding `Not verified` should change to `Verified`. _(This is unlike #44 where all messages updated based on one checkbox.)_

macro-verified

It will also make an API call to let the backend know that one macro verification step is complete. Each of these should be a persistence point.

### LINKED ISSUE #83: Expand Detail Component
I think the detail component can support details for syntax, validity, and quality, while also being used for the entire macro table. Header names will be picked up from the keys of the edit object.

wpears:
- Reworks EditsSyntacticalValidityDetail into a more general EditsDetail
- Applies EditsDetail to create EditsMacro
- Create EditsDetailRow which handles row level state for justification (prototype UI)

Closes #45, Closes #83


src/js/EditsContainer.jsx
```diff
@@ -2,29 +2,31 @@ var React = require('react');
...
 
 var EditsContainer = React.createClass({
   getInitialState: function() {
     return {
-    	"syntactical": [],
-      "validity": []
+      edits: {
+        syntactical: [],
+        validity: [],
+        macro: {}
```
awolfe76, line 14: Any reason `macro` is an object? We should be consistent one way or the other.

awolfe76:
@wpears - all reviewed. Theres the `{}` vs `[]` question but I have a few other questions, maybe we create a few more tickets?
- details for each macro edit 
  macro-detail
  - We should think about whether this detail info is in the intital api call or requires another. Would be great if it was there from the start. This would change the `edits.json` but only for the macro edits.
- add in the drop down for the justification, along with the text box
- maybe add a link to `#macro` in `InstitutionStatus.jsx`, looks like `case 3:` would work.

src/js/EditsContainer.jsx
wpears, line 14: Syntactical and validity are arrays because there are many different loan numbers which each have their own set of edits. For macro, however there is just one set of edits. This could be tweaked either way if we wanted everything to be either an array/object.. but I think this might be even more explicit that the data is _different_.

Additionally re: data structures, we may need to sit down and figure out all the ways we'll need to use the data so that it is stored in a format that allows the operations efficiently (if we need to look up every set of syntax edits by loan ID, perhaps rather than an arrray, we'd just store the data as an object with loanIDs as keys and edit arrays as values (or objects which contain the edit arrays and other stuff we don't yet have).

awolfe76, line 14: Cool on the edits ... just want to keep the backend folks in mind too. Might not matter at all for them though.

I agree on the data. We have a good idea about it but not the final. I kind of think we're learning a lot as we move along which could keep leading us toward the final. I feel better about that approach rather than trying to just figure it out. Building the app out has been a great way to keep learning what we really need.

awolfe76:
Any thoughts on creating more issues to add the couple things above or do you want to do them now?

wpears:
I'll create more issues, though I'm a little unclear on how the macro detail would be different than the description.

PR #91, opened 2016-04-14
Expandables

awolfe76:
- add cf-expandables to allow filers to show/hide the details of the edits returned
- closes #80

PR #92, opened 2016-04-14
NPM updates

awolfe76:
Depends on #91 
- followed instructions from [GHE]/gist/adamscott/4939082369fb422f6606
- updated dependencies
- fixed a warning due to using `this.props.key`, passing an id instead


wpears:
@awolfe76 snyk was failing on a pinned browser-sync dependency.... so I removed browser-sync ;)

awolfe76:
:+1: I don't think it was working correctly anyway ... maybe we can find another way at some point.

PR #103, opened 2016-05-11
Api module + expanded route support

wpears:
Adds on more routing goodies to #99.

Also adds a rudimentary module which shall be the gateway to the API (to further isolate changes which will crop up here).

Will support soon-to-arrive submission and edit component changes


src/js/AppContainer.jsx
```diff
@@ -6,19 +6,20 @@ var AppContainer = React.createClass({
...
+    var params = this.props.params || {};
 
     return (
       <div className="AppContainer">
-        <HomeLink/>
-        <UserHeading institution={{name: params.encodedInstitutionName}} year="2017" user={this.state.name}/>
-        {this.props.children}
+        <HomeLink year={params.year}/>
+        <UserHeading year={params.year} user={this.state.name}/>
+        {React.cloneElement(this.props.children, this.state.institution?{institution:this.state.institution}:null)}
```
awolfe76, line 22: Just want to be sure here, so rather than having all `props` this will now only give us the `institution`?

awolfe76:
Other than my one question should be good.

src/js/AppContainer.jsx
wpears, line 22: Actually before no props were being passed, now the institution is when it is present

awolfe76, line 22: 👍

Issue #98, opened 2016-05-02
Document API  at /api


wpears:
Return a json explanation of the various api endpoints (based on cfpb/hmda-platform#261) so that the API is discoverable

wpears:
This might actually be the wrong approach. We may not want to split the documentation between `api.md` and the API, and perhaps `/api` is better suited to be a simple pulse-check, general info endpoint (perhaps with a link to `api.md`).  Closing for now, will perhaps reopen/open a new issue once something is ticketed at hmda-platform

Issue #87, opened 2016-04-08
Rework router for hash fragments


wpears:
Redo the router to link to hash fragements and all links to the new macro/syntax/validity components where appropriate.

Created from #85

awolfe76:
Closing. Can be done in #256.

PR #105, opened 2016-05-12
Submissions routing

wpears:
Expands the api, dev-server and router to support handling submissions properly.  Mocked data is in place so that you should be able to route around the app without running into trouble.

Also includes a slight api refactor.

Doesn't include edit components yet, so parts of the app that display edits will look broken.


awolfe76:
Testing give a few warnings:

```
Warning: Failed propType: Required prop `year` was not specified in `Resubmit`. Check the render method of `InstitutionStatus`.
Warning: Failed propType: Required prop `year` was not specified in `Resubmit`. Check the render method of `InstitutionStatus`.
Warning: Failed propType: Required prop `id` was not specified in `Resubmit`. Check the render method of `InstitutionStatus`.

Warning: Failed propType: Required prop `submission` was not specified in `Resubmit`.
```

I _think_ this is because `__tests/InstitutionStatus.js` needs updated to pass those props.

awolfe76:
Same sort of warning for `HomeLink`:

```
Warning: Failed propType: Required prop `year` was not specified in `HomeLink`.
```

Again, I _think_ `__tests__/HomeLink.js` just needs updated to pass the props.

src/js/InstitutionStatus.jsx
```diff
@@ -8,17 +9,27 @@ var InstitutionStatus = React.createClass({
     institution: React.PropTypes.object.isRequired
```
awolfe76: Should `year: React.PropTypes.string.isRequired` be here too?

awolfe76:
I know its not in this PR, meant to mention it in the last one, but are we removing the ability to link to anchors on the page? Meaning like directly to an edit type, [`case 7:`](https://github.com/cfpb/hmda-platform-ui/pull/105/files#diff-40fdffa45f66992aa317e6e0349cd4d2R43).

awolfe76:
Might not be a thing at all yet (and I could be wrong) but when I view it locally odd things happen. Like the status for Bank 4 links to `http://localhost:3000/2017/bank4id/2` but that bank in the `submissions.json` file has 3 submissions. Shouldn't we be linking to the latest one in this case? _(might be a timestamp issue)_

src/js/api.js
```diff
@@ -8,23 +8,55 @@ function makeResponder(cb){
...
+function postHandler(url, cb, suffix){
+  if(typeof url === 'function'){
+    cb = url;
+    url = makeUrl(parseLocation, suffix);
+  }
+
+  superagent.post(url).end(makeResponder(cb));
+}
+
+function makeUrl(obj, suffix){
```
awolfe76, line 29: Can we make it more obvious that `makeUrl` is the only function exported here? Placement? Naming?

server/json/user1-institutions.json
```diff
@@ -6,47 +6,53 @@
       "status": {
         "code": 0,
         "message": ""
-      }
+      },
+      "currentSubmission": 0
```
awolfe76, line 10: Do we need `currentSubmission`? Can't we use the `timestamp` in `submissions.json`?

awolfe76:
Last thing ... I think it might be a good time to add a few things to the `README`.
- what the cases mean in `InstitutionStatus.jsx` and how they correspond to the FIs in the json
- a bit about the API; it's flow and what the json files are representing
- update the build and view process

src/js/api.js
wpears, line 29: Err you mean only function outside of the `get` and `post` functions? It has it's own entry in `module.exports`, see https://github.com/cfpb/hmda-platform-ui/pull/105/files#diff-3d4fce1ac7e3fbf5016b8881ccffd7bcR61

server/json/user1-institutions.json
wpears, line 10: We don't use the timestamp at all for logic, it is just there for info purposes. This way we don't have to think about time or syncing or timezones or anything. Just a backend service that increments the submission count.

wpears:
Good catch on the tests, probably shouldn't have ignored the red text just because they passed ;) Updating code now.

Anchors aren't precluded/eliminated as a possibility, but just haven't been included here. Anchors don't touch the backend and would just be a hint to the front-end to scroll down the page, so there isn't much complexity there or real difficulty in implementing. https://github.com/cfpb/hmda-platform-ui/issues/87 still stands.

re submission ids, the currentSubmission is the one that is shown when you click the view filing, etc. The reason the `submissions.json` file has an additional submission is so that the resubmit link works (which will increment currentSubmission, so the next submission needs to exist to be displayed).

Api documenting issue here: https://github.com/cfpb/hmda-platform-ui/issues/98
States documentation now ticketed at: https://github.com/cfpb/hmda-platform-ui/issues/106

src/js/api.js
awolfe76, line 29: Yeah, sorry. 😦 
I meant showing a difference between private and public. So the difference between `makeUrl` and `makeResponder`, `getHandler`, etc.

server/json/user1-institutions.json
awolfe76, line 10: Good point on the timestamp. Couldn't we also use the largest (if its incremental) `id` from the `submissions.json`?

wpears, line 10: Yeah, but the reason this is here is so that the API doesn't have to be hit for every institution to check what the submission status is. We certainly _could_ do that, but having the current submission associated with the institution is a convenience that will reduce backend load/front-end render times without (imo) mucking up the data model.

awolfe76, line 10: Sorry, I might be missing something else here. But the status is already there right? In the 

```
"status": {
  "code": 0,
  "message": ""
}
```

Anyway, I'll merge this and we can dig more with the backend crew when the API doc is ready.

Issue #106, opened 2016-05-13
Document states in the README


wpears:
see cfpb/hmda-platform#264

wpears:
Closed by #109

PR #109, opened 2016-05-17
Api docs

wpears:
Documents the API, as implemented in the dev server.

Also, has a few code changes where the implementation wasn't conforming to the spec as conceived.

### DIFF
--- a/api.md
+++ b/api.md
@@ -1,0 +1,73 @@
+* `/api`
+    * `GET` - Information about the API (version, metadata, links to repo, supported endpoints) - TODO
+
+* `/api/years`
+    * `GET` - List of all supported years. Structure: `{"years": ["2017"]}`
+
+* `/api/years/<year>`
+    * `GET` - Detailed attributes/metadata for a given year. Structure: [years](https://github.com/cfpb/hmda-platform-ui/blob/c7590b1e6bd3243a8cd7909463ab8e2dc8c046a5/server/json/year.json)
+
+* `/api/years/<year>/institutions`
+    * `GET` - Complete list of institutions a user can/could file for a given year. Structure: [institutions](https://github.com/cfpb/hmda-platform-ui/blob/c7590b1e6bd3243a8cd7909463ab8e2dc8c046a5/server/json/user1-institutions.json)
+
+* `/api/institutions/`
+    * `GET` - Complete list of institutions a user can/could file for *across all years*. Structure: [institutions](https://github.com/cfpb/hmda-platform-ui/blob/c7590b1e6bd3243a8cd7909463ab8e2dc8c046a5/server/json/user1-institutions.json), separated by year.
+
+* `/api/institutions/<institution>`
+    * `GET` - Data for a single institution *across all years*. Structure: [institution](https://github.com/cfpb/hmda-platform-ui/blob/c7590b1e6bd3243a8cd7909463ab8e2dc8c046a5/server/json/user1-institutions.json#L48-L56) collected in array based on id (`{institutions: [...]}`).
+
+* `/api/institutions/<institution>/years`
+    * `GET` - List of all years that a given institution filed/will file for. Structure: `{"years": ["2017"]}`
+  
+* `/api/years/<year>/institutions/<institution>`
+* `/api/institutions/<institution>/years/<year>`
+    * `GET` - Data for a single institution for a given year. Structure: [institutions](https://github.com/cfpb/hmda-platform-ui/blob/c7590b1e6bd3243a8cd7909463ab8e2dc8c046a5/server/json/user1-institutions.json#L48-L56)
+
+* `/api/years/<year>/institutions/<institution>/submissions`
+* `/api/institutions/<institution>/years/<year>/submissions`
+    * `GET` - List of all submissions for an institution in a given year. Structure: [submissions](https://github.com/cfpb/hmda-platform-ui/blob/c7590b1e6bd3243a8cd7909463ab8e2dc8c046a5/server/json/submissions.json#L26-L45) (without the extraneous key, which is just a convenience used by the dev server)
+    * `POST` - Create a new submission, incrementing the submission count. Return structure: `{"currentSubmission": 4}`
+
+* `/api/years/<year>/institutions/<institution>/submissions/<submission>`
+* `/api/institutions/<institution>/years/<year>/submissions/<submission>`
+    * `GET` - Data of a submission for an institution in a given year. Structure: [submissions](https://github.com/cfpb/hmda-platform-ui/blob/c7590b1e6bd3243a8cd7909463ab8e2dc8c046a5/server/json/submissions.json#L36-L45)
+    * `POST` - Endpoint for upload of LAR file. POST data: LAR file. Return structure: `{"progress": <progress endpoint>}`
+
+* `/api/years/<year>/institutions/<institution>/submissions/<submission>/progress`
+* `/api/institutions/<institution>/years/<year>/submissions/<submission>/progress`
+    * `GET` - Returns information about the progress of validation and submission in general. *This endpoint may support long-polling*. Structure:
+    ```
+    {
+      status: {
+        code: 3,
+        message: ""
+      },
+      editCounts: {
+        syntactical: 0,
+        validity: 1,
+        quality: 8,
+        macro: 0
+      }
+    }
+    ```
+
+* `/api/years/<year>/institutions/<institution>/submissions/<submission>/edits`
+* `/api/institutions/<institution>/years/<year>/submissions/<submission>/edits`
+    * `GET` - List of all edits for a given submission, grouped by edit type. Structure: `{"syntactical": {...}, "validity": {..}, "quality": {...}, "macro": {...}}`
+
+* `/api/years/<year>/institutions/<institution>/submissions/<submission>/edits`
+* `/api/institutions/<institution>/years/<year>/submissions/<submission>/edits`
+    * `GET` - List of all edits for a given submission, grouped by edit type. Structure: `{"syntactical": {...}, "validity": {..}, "quality": {...}, "macro": {...}}`
+    * `PUT` - Bulk update of edits, used for verification - TODO
+
+* `/api/years/<year>/institutions/<institution>/submissions/<submission>/syntactial|validity|quality|macro`
+* `/api/institutions/<institution>/years/<year>/submissions/<submission>/syntactial|validity|quality|macro`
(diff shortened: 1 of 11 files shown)


api.md
```diff
@@ -0,0 +1,73 @@
...
+      editCounts: {
+        syntactical: 0,
+        validity: 1,
+        quality: 8,
+        macro: 0
+      }
+    }
+    ```
+
+* `/api/years/<year>/institutions/<institution>/submissions/<submission>/edits`
```
awolfe76, line 54: This looks the same as below (lines 58-61). I think keep the one below because it has the `PUT`.

api.md
```diff
@@ -0,0 +1,73 @@
...
+* `/api/years`
+    * `GET` - List of all supported years. Structure: `{"years": ["2017"]}`
+
+* `/api/years/<year>`
+    * `GET` - Detailed attributes/metadata for a given year. Structure: [years](https://github.com/cfpb/hmda-platform-ui/blob/c7590b1e6bd3243a8cd7909463ab8e2dc8c046a5/server/json/year.json)
+
+* `/api/years/<year>/institutions`
+    * `GET` - Complete list of institutions a user can/could file for a given year. Structure: [institutions](https://github.com/cfpb/hmda-platform-ui/blob/c7590b1e6bd3243a8cd7909463ab8e2dc8c046a5/server/json/user1-institutions.json)
+
+* `/api/institutions/`
```
awolfe76: Just curious what the use case is for this list?

api.md
```diff
@@ -0,0 +1,73 @@
...
+
+* `/api/years/<year>/institutions/<institution>/submissions/<submission>/progress`
+* `/api/institutions/<institution>/years/<year>/submissions/<submission>/progress`
+    * `GET` - Returns information about the progress of validation and submission in general. *This endpoint may support long-polling*. Structure:
+    ```
+    {
+      status: {
+        code: 3,
+        message: ""
+      },
```
jmarin, line 44: What are `code` and `message` for this status?. Are the codes a finite list? If so, have we specified it somewhere else?

api.md
```diff
@@ -0,0 +1,73 @@
...
+    * `GET` - Complete list of institutions a user can/could file for a given year. Structure: [institutions](https://github.com/cfpb/hmda-platform-ui/blob/c7590b1e6bd3243a8cd7909463ab8e2dc8c046a5/server/json/user1-institutions.json)
+
+* `/api/institutions/`
+    * `GET` - Complete list of institutions a user can/could file for *across all years*. Structure: [institutions](https://github.com/cfpb/hmda-platform-ui/blob/c7590b1e6bd3243a8cd7909463ab8e2dc8c046a5/server/json/user1-institutions.json), separated by year.
+
+* `/api/institutions/<institution>`
+    * `GET` - Data for a single institution *across all years*. Structure: [institution](https://github.com/cfpb/hmda-platform-ui/blob/c7590b1e6bd3243a8cd7909463ab8e2dc8c046a5/server/json/user1-institutions.json#L48-L56) collected in array based on id (`{institutions: [...]}`).
+
+* `/api/institutions/<institution>/years`
+    * `GET` - List of all years that a given institution filed/will file for. Structure: `{"years": ["2017"]}`
```
awolfe76, line 20: I think we can remove the `/will file for`. I don't think we would know that information until a new panel is created.

awolfe76:
We will definitely keep updating this as we implement on the backend, but I think @jmarin @hkeeler @schbetsy @lgdean @nickgrippin should all take at least a quick look here.

Are we going to include the API version in the URL? Or is that covered by `/years/<year>`? (Had the same thought about grasshopper too.)

api.md
```diff
@@ -0,0 +1,73 @@
...
+    * `GET` - List of all supported years. Structure: `{"years": ["2017"]}`
+
+* `/api/years/<year>`
+    * `GET` - Detailed attributes/metadata for a given year. Structure: [years](https://github.com/cfpb/hmda-platform-ui/blob/c7590b1e6bd3243a8cd7909463ab8e2dc8c046a5/server/json/year.json)
+
+* `/api/years/<year>/institutions`
+    * `GET` - Complete list of institutions a user can/could file for a given year. Structure: [institutions](https://github.com/cfpb/hmda-platform-ui/blob/c7590b1e6bd3243a8cd7909463ab8e2dc8c046a5/server/json/user1-institutions.json)
+
+* `/api/institutions/`
+    * `GET` - Complete list of institutions a user can/could file for *across all years*. Structure: [institutions](https://github.com/cfpb/hmda-platform-ui/blob/c7590b1e6bd3243a8cd7909463ab8e2dc8c046a5/server/json/user1-institutions.json), separated by year.
```
jmarin, line 14: How do we get the user? This endpoint needs user information or else it would return the whole panel. I guess it's implied that user information will be delivered inside the request (headers?)

api.md
awolfe76, line 44: I _think_ the codes are from here:
https://github.com/cfpb/hmda-platform/issues/264#issuecomment-203686220

And the message would exist if the code is null. An example [here](https://github.com/cfpb/hmda-platform-ui/blob/master/server/json/submissions.json#L31). There are a few others in the file as well.

@wpears

api.md
```diff
@@ -0,0 +1,73 @@
...
+* `/api/years/<year>/institutions/<institution>/submissions/<submission>/edits`
+* `/api/institutions/<institution>/years/<year>/submissions/<submission>/edits`
+    * `GET` - List of all edits for a given submission, grouped by edit type. Structure: `{"syntactical": {...}, "validity": {..}, "quality": {...}, "macro": {...}}`
+
+* `/api/years/<year>/institutions/<institution>/submissions/<submission>/edits`
+* `/api/institutions/<institution>/years/<year>/submissions/<submission>/edits`
+    * `GET` - List of all edits for a given submission, grouped by edit type. Structure: `{"syntactical": {...}, "validity": {..}, "quality": {...}, "macro": {...}}`
+    * `PUT` - Bulk update of edits, used for verification - TODO
+
+* `/api/years/<year>/institutions/<institution>/submissions/<submission>/syntactial|validity|quality|macro`
```
awolfe76, line 63: Seeing it now, I'm not sure I like the idea of offering multiple endpoints for the same data. Seems to add confusion, and could be confusing for others as well. I think one clean path to the endpoint is enough. Same for the other places this exists. Anyone else?

api.md
jmarin, line 44: `null` is the root of all evil. I'd suggest using something else. To clarify, in `Scala` `nulls` are usually avoided (they are all over the place in `Java`). A `NullPointerException` is usually very hard to debug. Using `null` as a valid value is problematic, because we don't know if it's a missing value, or if something failed and the value could not be obtained.

wpears, line 44: Andrew is right about where the codes are coming from, and there is an open issue to document this https://github.com/cfpb/hmda-platform-ui/issues/106.

re: code being null; if that is a problem in Scala, we can use `-1` instead and just adjust the handling code slightly. Sound okay?

api.md
wpears, line 20: I mean for the current year. Perhaps a better phrase is `has filed or is filing for`

api.md
wpears, line 14: Yep, all the endpoints assume auth is handled separately and the users can be derived from headers.

api.md
wpears, line 63: The way I see it, the hierarchy here is a way to filter the results in different ways, but the order of filters doesn't necessarily matter. Say you have 6 marbles:

```
cats-eye boulder
cats-eye peewee
blue marble
blue boulder
speckled peewee
speckled marble
```

Whether you select first by size, second by color or first by color, second by size, you'll end up with the same group of marbles at the end.

As far as API usage goes, there are some scenarios that might make the client code a bit nicer if they can be flexible in the URIs they use.

If we want to keep the hierarchy strict, though, I'm okay with that, thought I'd prefer year first 8D

awolfe76, line 63: > As far as API usage goes, there are some scenarios that might make the client code a bit nicer if they can be flexible in the URIs they use.

Yeah, I agree. But for round 1 I think we're fine implementing only one and seeing if the need comes up. Might help @jmarin implement a little quicker. And yes, with the year first.

awolfe76:
This jumps off from [this line comment](#discussion_r63543312), but it started to get a bit large so pulled it out to here. 😄 

If possible, I think maybe the 2 versions we support should be with the `/years/<years>` but also without. Without always being the current year.

Right now, you can see in [`api.js`](https://github.com/cfpb/hmda-platform-ui/blob/master/src/js/api.js) we're picking apart and piecing together a URL based on the current URL. For example, if the URL is `cf.gov/hmda/filing/2017` we would be making calls to `/years/2017`. This is fine, but I think the better URL for the app in a current filing year would be without the `/2017`, the root of the app itself should live at `cf.gov/hmda/filing/`. (The app actually builds locally this way too, into a `/dist/<year>` directory.)

While not a huge lift for the filers, I have a feeling they would prefer to bookmark a URL to be able to go directly there and see the current app to be able to file ... without having to update the year. 

Supporting the API without the `/year/<year>` allows for this a bit more easily. So if filing for the current year the URL would be `cf.gov/hmda/filing/` and we would use the API **without** the `/year/<year>`. But if doing a resubmission the URL would be `cf.gov/hmda/filing/<year_of_resubmission>` (which could possibly be a whole different app) and we would use the corresponding API.

As an example, these lines I'm commenting on would be:
`/api/years/<year>/institutions/<institution>/submissions/<submission>/syntactial|validity|quality|macro`
to get a given year. And
`/api/institutions/<institution>/submissions/<submission>/syntactial|validity|quality|macro`
to get the current year.

Or, keeping the app URL the same, we could just default the API call to the current year. That would keep the API with the `/years/year`.

api.md
jmarin, line 44: Yeah, -1 would work. But I also wonder if code/message are really required. Would having something descriptive be better?. For instance:

``` javascript
{
  status: "Uploaded",
  editCounts: {
    syntactical: 0,
    validity: 1,
    quality: 8,
    macro: 0
  }
}
```

Just a thought. Would like to hear what you guys think

awolfe76, line 44: So maybe we add a few more codes for the possible errors, and terms to go with them, to https://github.com/cfpb/hmda-platform/issues/264#issuecomment-203686220 but only return the friendlier term in the response?

awolfe76:
I think that this is ok to merge now. Plenty of discussion going on but at a few different points so it's getting hard to follow.

I suggest we start to break this down into separate API issues on the hmda-platform repo and we can continue the separate discussions within those.

I also think we should start with the bare minimum to make the UI work in its current state:
- `/api/years/<year>/institutions`
- `/api/years/<year>/institutions/<institution>`
- `/api/years/<year>/institutions/<institution>/submissions`
- `/api/years/<year>/institutions/<institution>/submissions/<submission>`
- `/api/years/<year>/institutions/<institution>/submissions/<submission>/edits`
- `/api/institutions/<institution>/years/<year>/submissions/<submission>/syntactial|validity|quality|macro` _(used by /edits)_

Thoughts?

jmarin:
Works for me

PR #110, opened 2016-05-19
Signature

### LINKED ISSUE #50: Component for confirmation and signature
This component will render high level data about the submission as a way to verify the respondent and file information are correct. It will also contain a checkbox and verification message which acts as the final step in the submission process.

confirm

Respondent data required:
- Name
- ID
- Tax ID
- Agency
- Contact name
- Phone
- Email

File data required:
- File name
- Year
- Total Loans/Applications

awolfe76:
**Needs tests**
Adds in the singature.jsx and makes a few other minor updates to render within the layout.

closes #50

### DIFF
--- a/src/js/InstitutionStatus.jsx
+++ b/src/js/InstitutionStatus.jsx
@@ -52,6 +52,7 @@ var InstitutionStatus = React.createClass({
         break;
       case 9:
       case 10:
+        statusLink = <Link to={appLink}>Sign your filing</Link>
         statusText = <p>All checks and verification complete. Review the summary report and sign the submission.</p>
         break;
       case 11:
--- a/src/js/SubmissionContainer.jsx
+++ b/src/js/SubmissionContainer.jsx
@@ -2,6 +2,7 @@ var React = require('react');
 var api = require('./api');
 var UploadForm = require('./UploadForm.jsx');
 var EditsContainer = require('./EditsContainer.jsx')
+var Signature = require('./Signature.jsx');
 
 function uploadCb(){
   console.log('submited, -> transition');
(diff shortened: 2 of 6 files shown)


src/js/SubmissionContainer.jsx
```diff
@@ -52,22 +53,26 @@ var SubmissionContainer = React.createClass({
       )
     }
 
-    if(code > 3) progress = <p>Progress component goes here</p>
+    if(code > 3) progress = <p className="third">Progress component goes here</p>
 
-    if(code > 6) editsContainer = <EditsContainer/>
+    if(code > 5) editsContainer = <EditsContainer/>
```
awolfe76, line 58: I changed this because `6: Validation begins`, so wouldn't we want to start to be able to show edits right away?

wpears, line 58: Ideally, yes, but that implies we'll be streaming back the edit results rather than just the edit counts which I think we're avoiding for now.

Personally I think that getting as much data as soon as possible is a good thing, but it really depends on the speed of validation. If validation takes many minutes for large FIs, being able to investigate errors immediately I think is a usability win. But the extra streaming infrastructure might not be a good investment if validation takes < 20 seconds for 99% of users.

So for now I say we stick with only showing edits when validation has complete.

src/js/InstitutionStatus.jsx
```diff
@@ -52,6 +52,7 @@ var InstitutionStatus = React.createClass({
         break;
       case 9:
       case 10:
+        statusLink = <Link to={appLink}>Sign your filing</Link>
```
wpears, line 55: Don't we still want this link here? Since we still need to open the submission container view to do the actual signing.

src/js/SubmissionContainer.jsx
awolfe76, line 58: Ah, yeah. Right.

What do you think about adding a few more states?
- `6`: Syntactical begins
- `7`: Syntactical completes
- `8`: Validity begins
- `9`: Validity completes

For each edit type.

Then, when polling, we could start to show by type?

src/js/InstitutionStatus.jsx
awolfe76, line 55: Not sure I understand. Isn't the link there in `case 10`?

wpears, line 55: Huh. It sure is. My brain must be broken, nevermind ^^;

src/js/SubmissionContainer.jsx
wpears, line 58: I don't know if we can guarantee the order of edits, so this might be tenuous.

src/js/InstitutionStatus.jsx
awolfe76, line 55: 😄

src/js/SubmissionContainer.jsx
awolfe76, line 58: At one point we had discussed not running quality and macro until syntactical and validity were complete and correct. If that were possible then we should be ok?

@jmarin - thoughts on adding those extra states to be able to show some edits as those edit type checks complete?

awolfe76:
@wpears - seems like a test for this is really a test against React itself, at least as it stands.
We should, in some other issue, write test(s) against `SubmissionContainer.jsx` though.
What do you think?

wpears:
I'll buy it for now ;)

One we actually put some functionality in the signature component we'll add some useful tests.

PR #111, opened 2016-05-23
API to institution-first only

wpears:
Updates the proposed api to only include institution-first style.

"Year" has been renamed to "period" across the project.

Components have been updated to handle new institution-first style (and front-end routing has been updated as the routing is also institution-first).


awolfe76:
👍

Issue #46, opened 2016-03-16
Component to render institutional register summary (IRS)


awolfe76:
The IRS displays "all MSA/MDs where my institution has a home or branch office (and took loan/applications in that office) are listed on the IRS. Each MSA/MD listed is an MSA/MD in which we have a home or branch office." The exact content for this section TBD, but we need a place for it.

Also, this section "Please review each of the **2** MSA/MDs listed below. If you disagree please correct and re-upload the updated file" requires the MSA/MD count.
### _With_ syntactical and validity errors

![irs](https://cloud.githubusercontent.com/assets/2932572/13812587/599d2a40-eb53-11e5-96b7-c1cf45eeed54.png)

Data required:
- MSA/MD ID
- MSA/MD name
- Total LARs
- Total amount ($ in thousands)
- # on CONV loan type
- # of FHA loan type
- # of VA loan type
- # of FSA/RHS loan type
- # of 1-4 family property type
- # of MFD property type
- # of Multi-family property type
- # of Home Purchase loan purpose
- # of Home Improvement loan purpose
- # of Refinance loan purpose
### _Without_ syntactical and validity errors

Once the syntactical and validity edits have run without error, this component renders a bit differently. It should then display a checkbox and verification message for the filer to verify the IRS.

![irs-verify](https://cloud.githubusercontent.com/assets/2932572/13812762/5cc33858-eb54-11e5-9c98-ecf5021e2809.png)

This checkbox will also make an API call to let the backend know this step is complete. This should be a persistence point.

PR #112, opened 2016-05-24
Start of IRS Report

awolfe76:
This is the start of the IRS Report page (see #46). A later PR will finish it, with tests. Wanted to send this PR first though because of the changes to the states.
- adds new states for IRS creation and verification
  - creation state needed because the report will be creating on-the-fly after all other verifications are complete
- creates generic `IRSReport.jsx` component
- changes to `.json` and `.jsx` files to demonstrate
- route updates and additions

### DIFF
--- a/src/js/SubmissionContainer.jsx
+++ b/src/js/SubmissionContainer.jsx
@@ -56,30 +56,27 @@ var SubmissionContainer = React.createClass({
       )
     }
 
-    if(code > 3) progress = <p className="third">Progress component goes here</p>
+    if(code > 3) progress = <p>Progress component goes here</p>
 
     if(code > 5) editsContainer = <EditsContainer/>
 
     if(code > 9){
       irs = <IRS/>;
-      uploadForm = null;
-      editsContainer = null;
     }
 
     if(code > 10){
       summary = <p>Summary component here</p>
       sign = <Signature/>
-      uploadForm = null;
-      editsContainer = null;
-      irs = null;
     }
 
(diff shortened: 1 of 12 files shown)


src/js/SubmissionContainer.jsx
```diff
@@ -57,11 +60,18 @@ var SubmissionContainer = React.createClass({
 
     if(code > 5) editsContainer = <EditsContainer/>
 
-    if(code > 8){
+    if(code > 9){
+      irs = <IRS/>;
+      uploadForm = null;
```
wpears: I don't think we want `null` here (or below if `code > 10`). At least how I've been thinking about it, we'll just keep adding new components onto the page as the user move through the process. This way they'll know precisely where they are in the process, etc.

wpears:
@awolfe76 Just the one comment above, otherwise seems good for the next PR

src/js/SubmissionContainer.jsx
awolfe76: Hmm, ok. I think I see what you mean. Basically just leave all the components there once they are added? And let them render however they'll render ... could be empty even but that's ok.

awolfe76:
Updates added. Mind giving it another quick look, at least at [`SubmissionContainer`](https://github.com/cfpb/hmda-platform-ui/pull/112/files#diff-9ceb5c7e53f49f44cd899504ec565cf1).

wpears:
🚀

Issue #116, opened 2016-05-26
CSS refactor


awolfe76:
A specific example can be seen [here](https://github.com/cfpb/hmda-platform-ui/pull/114#diff-for-comment-64652088).

But overall, the CSS can use a quick look through now that we have a better idea of the components.

Maybe after #47, #48, and #49. Those should give us most of the UI.

awolfe76:
Closing. Will use #155 do handle this instead.

Issue #115, opened 2016-05-26
Test for IRS component


awolfe76:
The `IRSReport.jsx` component needs better tests.
- check that the `onChange` function is updating the state in the parent (`SubmissionContainer.jsx`)
- check that the checkbox is toggling

PR #114, opened 2016-05-25
IRS Report

awolfe76:
closes #46 
- adds tests (also for signature)
- a few minor updates to use `self` instead of `_this`
- some css for the IRS table
- json structure for the IRS report

PR #117, opened 2016-05-26
IRS verified

awolfe76:
Adds in the more detailed description of the IRS report.
Adds in the checkbox for verifying (or un-verifying) the IRS report. When this happens a parent component function is called in order to update the state, specifically the status code. This allows us to show/hide the signature component. The signature component should NOT render if the IRS has not been verified.

I created #115 to update the tests for this. Wanted to get your thoughts on this approach first.

Also created #116 for the CSS (mentioned in #114).

### DIFF
--- a/src/js/SubmissionContainer.jsx
+++ b/src/js/SubmissionContainer.jsx
@@ -2,7 +2,7 @@ var React = require('react');
 var api = require('./api');
 var UploadForm = require('./UploadForm.jsx');
 var EditsContainer = require('./EditsContainer.jsx');
-var IRS = require('./IRSReport.jsx');
+var IRSReport = require('./IRSReport.jsx');
 var Signature = require('./Signature.jsx');
 
 function uploadCb(){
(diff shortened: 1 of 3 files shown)


src/js/SubmissionContainer.jsx
```diff
@@ -33,6 +33,26 @@ var SubmissionContainer = React.createClass({
     }
   },
 
+  toggleIRSCheck: function(e){
+    // should make an API call here to POST verification
+    // and if everything is ok it would return the new status
+    if (e.target.checked) {
```
wpears, line 39: I think state should flow entirely from the backend, meaning that in the API call here, the callback needs to verify that the status returned is equal to the state set below. I think setting the state outside of the callback is fine for usability reasons (immediately move to signature component rather than have to wait for callback) and since the backend will pretty much just be noting that a the check box was checked and advancing the state. Do note that a check will need to exist in the callback when it's implemented here. I'm going to ticket something here.

wpears:
Follow up with #118

PR #121, opened 2016-06-01
IRS `POST` API

### LINKED ISSUE #118: Front-/back-end to support IRS verification
Users need to be able to verify their IRS to move to the summary page. #117 supports this, but still needs an API route to post to and the associated front-end callback to verify that the state transition has been recorded.

awolfe76:
closes #118 

adds in the POST API for the IRS report
the POST API takes the current status as a parameter and simply swaps its value
10 and 11 are currently the only states where the IRS report is involved

### DIFF
--- a/server/routes/irs.js
+++ b/server/routes/irs.js
@@ -6,11 +6,18 @@ router.get('/', function (req, res) {
   res.send(irsObj);
 });
 
-router.post('/:irs', function (req, res){
-  console.log(req);
-  console.log(res);
+router.post('/:status', function (req, res){
+  var newStatus;
+  if (req.params.status === '10'){
+    newStatus = '11';
+  } else {
+    newStatus = '10';
+  }
   res.status(202).send({
-
+    status: {
+      code: newStatus,
+      message: ""
+    }
   });
 });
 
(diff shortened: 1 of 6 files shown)


server/routes/irs.js
```diff
@@ -6,4 +6,19 @@ router.get('/', function (req, res) {
...
+  var newStatus;
+  if (req.params.status === '10'){
+    newStatus = '11';
+  } else {
+    newStatus = '10';
+  }
+  res.status(202).send({
+    status: {
+      code: newStatus,
+      message: ""
```
awolfe76, line 15: We _could_ also update with a message to help the front-end. Probably not necessary, at least not yet.

wpears:
I think it might be better to POST an object like `{verified: true}` or `{verified: false}` then let the backend handle the status stuff. Otherwise, we'll be muddying our source of truth for the backend. Ideally, status should only be set in response to updates from the backend (with UX complications, like updating the state assuming the backend will not error and checking that that was indeed the case in a callback).

So that'd be a POST to `/api/institutions/<institution>/periods/<period>/submissions/<submission>/irs` with POST data for  `{verified: <bool>}` which the backend will then parse and use its internal state to make decisions.

wpears:
Also, we can now add tests and docs (in a later PR), ticketed in #115 and #124

awolfe76:
@wpears - i think this has it. Had to add `body-parser` to be able to get to `req.body`.

wpears:
👍

PR #123, opened 2016-06-01
restart express on /server update

awolfe76:
Helpful when working on the API.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "test": "jest",
     "snyk": "snyk test",
     "clean": "rimraf ./dist && mkdir -p ./dist ./dist/js ./dist/css ./dist/fonts",
-    "start": "./server/server.js",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -7,7 +7,7 @@
     "test": "jest",
     "snyk": "snyk test",
     "clean": "rimraf ./dist && mkdir -p ./dist ./dist/js ./dist/css ./dist/fonts",
-    "start": "./server/server.js",
+    "start": "nodemon ./server/server.js",
```
wpears: Does the nodemon need to be here? Shouldn't we just change the `watch` command to `npm run watch:server` instead of `npm start`? Also, you can change the `watch:server` command's `-x` flag to `'npm start'` (which runs the special script named start ^_~.. equivalent but slightly shorter)

awolfe76:
@wpears - I _think_ this does it.

Issue #97, opened 2016-04-29
Filter LARs from API returns based on external fields file


wpears:
Rather than requiring edits to return fields they are operating on, we should instead simply return affected LARs and allow the frontend to display just affected fields based on an external mapping ie

```
{
  "S028": ["loanID", "race"],
  "S035": ["race","sex"],
  ...
}
```

The frontend could keep this mapping in memory and use it to render just the affected fields for a given edit.

wpears:
This is actually better suited to the backend; namely, it's much better to only send required data over the wire. Created https://github.com/cfpb/hmda-platform/issues/429

PR #126, opened 2016-06-06
Edit components

wpears:
- Edit Components rewritten for new api
- Edit Components sortable by type
- expandables rewritten for performance/play nicely with react

### DIFF
--- a/src/js/EditsDetailRow.jsx
+++ b/src/js/EditsDetailRow.jsx
@@ -18,7 +18,7 @@ var EditsDetailRow = React.createClass({
       if(this.state.verified) return this.state.verification;
       else return <textarea onChange={this.updateText} value={this.state.verification}/>
     }
-    if(field === 'lar') return JSON.stringify(detail[field]);
+    if(field === 'lar') return detail[field].loanId;
     return detail[field];
   },
 
--- a/src/js/EditsDetailRow.jsx
+++ b/src/js/EditsDetailRow.jsx
@@ -1,9 +1,8 @@
 var React = require('react');
-
 var EditsDetailRow = React.createClass({
 
   propTypes: {
-    edit: React.PropTypes.object.isRequired,
+    detail: React.PropTypes.object.isRequired,
     id: React.PropTypes.number
   },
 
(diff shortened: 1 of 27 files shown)


src/js/EditsDetailRow.jsx
```diff
@@ -1,38 +1,38 @@
...
+  makeTdContent: function(detail, field){
     if(field === 'verification'){
-      if(this.state.verification) return this.state.verification;
-      else return <textarea onChange={this.updateText}value={this.state.verification}/>
+      if(this.state.verified) return this.state.verification;
+      else return <textarea onChange={this.updateText} value={this.state.verification}/>
     }
-    if(field === 'lar') return JSON.stringify(edit[field]);
-    return edit[field];
+    if(field === 'lar') return JSON.stringify(detail[field]);
```
awolfe76: When grouped by type, this is rendering `{"loanId":"s1"}`. Shouldn't we just render the loadID value?

src/js/api.js
```diff
@@ -64,10 +64,14 @@ module.exports = {
    return postHandler(url, cb, '/submissions');
  },
 
- getEdits: function(url, cb){
+ getEditsByType: function(url, cb){
    return getHandler(url, cb, '/edits');
  },
 
+ getEditsByRow: function(url, cb){
```
awolfe76, line 71: Does these need something added to `api.md`?

awolfe76:
@wpears - couple small comments/questions but 👍

src/js/EditsDetailRow.jsx
wpears: So this is actually where https://github.com/cfpb/hmda-platform-ui/issues/97 comes in, since there will be many fields but we're likely only interested in one to several per edit. I'll make the change here though as well, for great justice ;)

src/js/api.js
wpears, line 71: There actually is already an entry, but in checking it out I realized there is an error, so I'll fix that as well ;)

wpears:
Should be good to go ^^;

src/js/api.js
awolfe76, line 71: 💯

PR #130, opened 2016-06-09
Q029 component rendering

### LINKED ISSUE #47: Component for Q029 edit
The Q029 edit is when "MSA/MD = NA and state/county (when county is small and tract equals NA), or state/county/tract is a valid combination and is located completely in a MSA/MD." (This is the content used within the component.)
### _With_ syntactical and validity errors

If syntactical and validity errors still exist an informational message should display above the table. The message should read "To verify these edits all syntactical and validity edits must pass." (See #45 as an example.)

Also in the case where syntactical and validity errors still exist the table should render without the final column containing the dropdown. (See the example screenshot below.)
### _Without_ syntactical and validity errors

When there are _no_ syntactical and validity errors filers are allowed to verify the results using a simple dropdown with the options of:
- Yes, the listed MSA/MD should be updated
- No, the listed MSA/MD should not be updated

The dropdown is available for each row and also in the heading row. The heading row dropdown can be used to update all other rows.

q029

Data required:
- Loan number
- State code
- County code
- Census tract
- Recommended (tract that we recommend)

Each dropdown selection will make an API call to the backend to be saved. Each of these should be a persistence point.

### LINKED ISSUE #48: Component for Q595 edit
The Q595 edit is when "MSA/MD not on respondent panel." (This is the content used within the component.)
### _With_ syntactical and validity errors

If syntactical and validity errors still exist an informational message should display above the table. The message should read "To verify these edits all syntactical and validity edits must pass." (See #45 as an example.)

Also in the case where syntactical and validity errors still exist the table should render without the final column, including header, containing the dropdown. (See the example screenshot below.)
### _Without_ syntactical and validity errors

When there are _no_ syntactical and validity errors filers are allowed to verify the results using a simple dropdown with the options of:
- Yes
- No

The dropdown is available for each row.

q595

Data required:
- MSA/MD ID
- MSA/MD name
- MSA/MD LAR count

Each dropdown selection will make an API call to the backend to be saved. Each of these should be a persistence point.

awolfe76:
Closes #47. Closes #48.

_This PR does not deal with the `POST`. Created #136 and #137 to handle this._

This PR is ready. It:
- does the `GET` for Q029 and Q595
- updates the edits API to include Q029 and Q595 edits
- renders and tests
- updates API doc

Changes only one endpoint -
`/api/institutions/<institution>/periods/<period>/submissions/<submission>/edits/syntactial|validity|quality|macro`

To add `q029|q595` as options, along with the `syntactial|validity|quality|macro`.

### DIFF
--- a/api.md
+++ b/api.md
@@ -39,10 +39,10 @@
     ```
 
 * `/api/institutions/<institution>/periods/<period>/submissions/<submission>/edits`
-    * `GET` - List of all edits for a given submission, grouped by edit type. Structure: `{"syntactical": {...}, "validity": {..}, "quality": {...}, "macro": {...}}`
+    * `GET` - List of all edits for a given submission, grouped by edit type. Structure: `{"syntactical": {...}, "validity": {..}, "quality": {...}, "macro": {...}, "q029": {...}, "q595": {...}}`
     * `PUT` - Bulk update of edits, used for verification - TODO
 
-* `/api/institutions/<institution>/periods/<period>/submissions/<submission>/edits/syntactial|validity|quality|macro`
+* `/api/institutions/<institution>/periods/<period>/submissions/<submission>/edits/syntactial|validity|quality|macro|q029|q595`
     * `GET` - Edits for a given submission, of a single type. Structure: [syntactical](https://github.com/cfpb/hmda-platform-ui/blob/c7590b1e6bd3243a8cd7909463ab8e2dc8c046a5/server/json/syntactical.json) [validity](https://github.com/cfpb/hmda-platform-ui/blob/c7590b1e6bd3243a8cd7909463ab8e2dc8c046a5/server/json/validity.json) [quality](https://github.com/cfpb/hmda-platform-ui/blob/c7590b1e6bd3243a8cd7909463ab8e2dc8c046a5/server/json/validity.json) [macro](https://github.com/cfpb/hmda-platform-ui/blob/c7590b1e6bd3243a8cd7909463ab8e2dc8c046a5/server/json/macro.json)
 
 * `/api/institutions/<institution>/periods/<period>/submissions/<submission>/edits/lars`
--- a/server/json/q029.json
+++ b/server/json/q029.json
@@ -1,6 +1,6 @@
 {
   "type": "q029",
-  "lars": [
+  "edits": [
     {
       "lar": {"loanId": "00001"},
       "stateCode": "36",
--- a/server/json/q029.json
+++ b/server/json/q029.json
@@ -7,7 +7,7 @@
       "countyCode": "021",
       "censusTract": "0004.02",
       "recommendedMSA": "00",
-      "verification": ""
+      "verified": null
     },
     {
       "lar": {"loanId": "00002"},
@@ -15,7 +15,7 @@
       "countyCode": "021",
       "censusTract": "0004.02",
       "recommendedMSA": "00",
-      "verification": "yes"
+      "verified": true
     },
     {
       "lar": {"loanId": "00003"},
@@ -23,7 +23,7 @@
       "countyCode": "021",
       "censusTract": "0004.02",
       "recommendedMSA": "00",
-      "verification": "no"
+      "verified": false
     }
   ]
 }
--- a/server/json/q029.json
+++ b/server/json/q029.json
@@ -1,0 +1,29 @@
+{
+  "type": "q029",
+  "lars": [
+    {
+      "lar": {"loanId": "00001"},
+      "stateCode": "36",
+      "countyCode": "021",
+      "censusTract": "0004.02",
+      "recommendedMSA": "00",
+      "verification": ""
+    },
+    {
+      "lar": {"loanId": "00002"},
+      "stateCode": "36",
+      "countyCode": "021",
+      "censusTract": "0004.02",
+      "recommendedMSA": "00",
+      "verification": "yes"
+    },
+    {
+      "lar": {"loanId": "00003"},
+      "stateCode": "36",
+      "countyCode": "021",
+      "censusTract": "0004.02",
+      "recommendedMSA": "00",
+      "verification": "no"
+    }
+  ]
+}
--- a/src/js/EditsQ029.jsx
+++ b/src/js/EditsQ029.jsx
@@ -6,13 +6,15 @@ var EditsQ029 = React.createClass({
     group: React.PropTypes.array
   },
 
-  makeSelect: function() {
+  makeRadio: function(count, renderHeading) {
+    var heading = null;
+    if (renderHeading) heading = "The Listed MSA/MD should be updated.";
     return (
-      <select>
-        <option value="Select">Select ...</option>
-        <option value="Yes">Yes, the listed MSA/MD should be updated.</option>
-        <option value="No">No, the listed MSA/MD should not be updated.</option>
-      </select>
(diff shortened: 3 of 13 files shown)


server/json/q029.json
```diff
@@ -0,0 +1,29 @@
+{
+  "type": "q029",
+  "edits": [
+    {
+      "lar": {"loanId": "00001"},
+      "stateCode": "36",
+      "countyCode": "021",
+      "censusTract": "0004.02",
+      "recommendedMSA": "00",
+      "verification": ""
```
wpears: I think we probably want a different key other than verification and for the value to be a boolean. We could make verification a boolean, but I think that would complicate/muddle things.

wpears: So maybe "verified" which can be true or false

src/js/EditsQ029.jsx
```diff
@@ -0,0 +1,52 @@
...
+  propTypes: {
+    group: React.PropTypes.array
+  },
+
+  makeSelect: function() {
+    return (
+      <select>
+        <option value="Select">Select ...</option>
+        <option value="Yes">Yes, the listed MSA/MD should be updated.</option>
+        <option value="No">No, the listed MSA/mD should not be updated.</option>
```
wpears: typo on mD

src/js/EditsQ029.jsx
```diff
@@ -0,0 +1,52 @@
+var React = require('react');
+var EditsDetail = require('./EditsDetail.jsx');
+
+var EditsQ029 = React.createClass({
+  propTypes: {
+    group: React.PropTypes.array
+  },
+
+  makeSelect: function() {
```
wpears: Does this make more sense as a checkbox? A select for a yes/no seems weird to me

src/js/EditsQ595.jsx
```diff
@@ -0,0 +1,47 @@
+var React = require('react');
+
+var EditsQ595 = React.createClass({
+  propTypes: {
+    group: React.PropTypes.array
+  },
+
+  makeSelect: function() {
```
wpears: Same as 029, checkbox vs select

wpears:
I think both Q029 and Q595 should depend on component state that gets updated onClick. For example, right now the group by row/group by type button triggers a re-render which wipes out selected values.

src/js/EditsQ029.jsx
wpears: I think the select elements should actually be radios (with no default checked). See [this article](http://uxmovement.com/forms/why-users-abandon-forms-with-select-menus/) for some usability troubles with select.

src/js/EditsQ595.jsx
wpears: Same comment re: radio buttons

api.md
```diff
@@ -39,10 +39,10 @@
     ```
 
 * `/api/institutions/<institution>/periods/<period>/submissions/<submission>/edits`
-    * `GET` - List of all edits for a given submission, grouped by edit type. Structure: `{"syntactical": {...}, "validity": {..}, "quality": {...}, "macro": {...}}`
+    * `GET` - List of all edits for a given submission, grouped by edit type. Structure: `{"syntactical": {...}, "validity": {..}, "quality": {...}, "macro": {...}, "q029": {...}, "q595": {...}}`
     * `PUT` - Bulk update of edits, used for verification - TODO
 
-* `/api/institutions/<institution>/periods/<period>/submissions/<submission>/edits/syntactial|validity|quality|macro`
+* `/api/institutions/<institution>/periods/<period>/submissions/<submission>/edits/syntactial|validity|quality|macro|q029|q595`
```
wpears, line 45: So I have some thoughts/revisions on this structure:
- `.../edits/<edit>` should be an endpoint which returns just the `<edit>` specified. The endpoints for q029 and q595 would be captured by this.
- syntactical/validity/quality/macro/ should be part of a querystring with key 'type' so to get all macro edits you'd hit `.../edits?type=macro`
- a `lar` key in the querystring would allow filtering by lar, which could perhaps be expanded to accept a range of values for paging (stay tuned)
- a `groupBy` key in the querystring would support sorting based on lar where appropriate

I'll be working on these ideas in the next few days; out-of-scope for this PR; but thought I should mention it here so we're thinking about things the same way

server/json/q029.json
awolfe76: Cool with me.
In the case where they haven't verified either way should it be: `"verified": null`?

awolfe76:
> I think both Q029 and Q595 should depend on component state that gets updated onClick. For example, right now the group by row/group by type button triggers a re-render which wipes out selected values.

Right, I didn't bother with a `POST` of any kind in this PR. But ... this seems related to the flow you were mentioning in the meeting a few minutes ago. If that's right, can you push your branch up so I can see how you're handling this?

src/js/EditsQ029.jsx
awolfe76: This sounds ok to me. The `select` idea was taken from pilot. 
@lesgou @debseidner were there any reasons to use a dropdown vs radio buttons?

This could be fixed in a separate PR as well.

api.md
awolfe76, line 45: For the first 2 bullets:

> - .../edits/<edit> should be an endpoint which returns just the <edit> specified. The endpoints for q029 and q595 would be captured by this.
> - syntactical/validity/quality/macro/ should be part of a querystring with key 'type' so to get all macro edits you'd hit .../edits?type=macro

Are these doing the same thing? Or do you see the Q029 and Q595 being returned in the proper  group there too?

> - a lar key in the querystring would allow filtering by lar, which could perhaps be expanded to accept a range of values for paging (stay tuned)

Not sure I follow here. Do we have a need in the UI to do this filtering? "Show me a single LAR"?

src/js/EditsQ029.jsx
lesgou: @awolfe76 @wpears There is no specific reason to use the drop down. Radio button is good (as long as there are two)

api.md
wpears, line 45: First two bullets:
Querying for Q029 edits would involve hitting the endpoint at `.../edits/q029` while getting all quality edits would be `.../edits?type=quality`. Quality is a filter and the individual edits are members of the hierarchy.

re: lar key, this might be the wrong track, as we'll need to support PUTs on individual edits/lars see #138. So perhaps something more structured will be needed. `.../edits/<edit>/lars/<lar>`

server/json/q029.json
wpears: Hmm. I think it should default to false and you just make sure that the false radio isn't selected for the initial render (so people don't just breeze through the already selected buttons)

awolfe76: But what about the case where they leave and come back? That initial render could have some `true`, some `false`, and some they haven't selected yet. If we're returning `false` for those we wouldn't know the difference.

awolfe76:
@wpears I think this is ready except for this comment

> But what about the case where they leave and come back? That initial render could have some true, some false, and some they haven't selected yet. If we're returning false for those we wouldn't know the difference.

server/json/q029.json
wpears: By default to false, I mean they'll default to false in the data. I'm working on this code, so you'll see what I mean shortly ;)

awolfe76: Cool. I think this is ready to merge then.

PR #132, opened 2016-06-09
Validation progress component

wpears:
Implements a validation progress component and supporting backend/api functions.

Currently polls every 500ms (after each response waits 500ms then makes another call). This should likely be larger (probably 5s), but it's nice for testing app flow now. Could be set based on environment, but doesn't need too much smarts for the POC phase.

### DIFF
--- a/src/js/ValidationProgress.jsx
+++ b/src/js/ValidationProgress.jsx
@@ -25,7 +25,6 @@ class ValidationProgress extends React.Component {
     var validationStarted = null;
     var validationComplete = null;
     var code = this.state.statusCode;
-    console.log(code, 'rendering');
 
     if(code > 2) uploadComplete = <p>Upload complete</p>;
     if(code > 3) parsingStarted = <p>Parsing started...</p>;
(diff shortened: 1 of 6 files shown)


src/js/ValidationProgress.jsx
```diff
@@ -0,0 +1,50 @@
+var React = require('react');
+var api = require('./api');
+
+class ValidationProgress extends React.Component {
+
+  constructor(props){
```
awolfe76, line 6: Any reason you went the ES6 route?

awolfe76:
Just the one question. Other than that looks good!

src/js/ValidationProgress.jsx
wpears, line 6: Mostly for fun; sharpen the saw and all that. If you'd prefer to keep it strictly ES5 we can strip it out; I just figure that since we already are leaning on babel, we might as well get some mileage from it ;)

awolfe76, line 6: Honestly, I like the idea. Just threw me for a minute. 😕 
I create #133. I think we should stick to a path but think ES6 is the correct one. I think we even talked about that before we started to build.

PR #135, opened 2016-06-10
Patch validationprogress component to wait for mounting to setState

### LINKED ISSUE #134: Warnings in the UI
@wpears, after #132 I'm noticing a few warnings in the UI - 

`app.min.js:1510 Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the EditsContainer component.`

`app.min.js:1510 Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the IRS component.`

It does prevent the IRS report from rendering, even though the edits do.

These happen when running master now. Sorry I didn't notice them earlier.

wpears:
Fixes #134

### DIFF
--- a/__tests__/ValidationProgress.js
+++ b/__tests__/ValidationProgress.js
@@ -32,33 +32,26 @@ describe('ValidationProgress', function(){
     setTimeout(wrappedCb, 0);
   });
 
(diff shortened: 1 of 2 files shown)


awolfe76:
👍

PR #142, opened 2016-06-15
Generalized callbacks / app flow

wpears:
Includes:
- error first callbacks (so errors aren't just swallowed by the api)
- updates to IRS and signature to use standard callback patterns
- expands and refactors the API
- allows transitioning to IRS report after verifying all quality/macro

Going to integrate this with the q029/q595 work tomorrow; but this kinda ballooned as I realized more stuff needed to be changed, so I wanted it in ASAP

### DIFF
--- a/__tests__/Signature.js
+++ b/__tests__/Signature.js
@@ -9,7 +9,7 @@ var fs = require('fs');
 var api = require('../src/js/api');
 
 api.postSignature = jest.fn(function(cb){
-  cb({
+  cb(null, {
     status: {
       code: state,
       message: ""
--- a/__tests__/Signature.js
+++ b/__tests__/Signature.js
@@ -5,34 +5,34 @@ var ReactDOM = require('react-dom');
 var TestUtils = require('react-addons-test-utils');
 
 var Signature = require('../src/js/Signature.jsx');
-var fs = require('fs');
 var api = require('../src/js/api');
 
+var code = 13;
 api.postSignature = jest.fn(function(cb){
   cb(null, {
     status: {
-      code: state,
-      message: ""
+      code: code,
+      message: ''
     },
     timestamp: Date.now(),
-    receipt: 'somehash'
+    receipt: 'dc9e5827abb678f54103e7b89435abf9b36648797ebb6516a52ab33ab4e46cee'
   });
+  code = code === 13 ? 12 : 13;
 });
 
 describe('irs report', function(){
-  var changeHandlerTrue = function(e){
-    expect(e.target.checked).toBeTruthy();
+  var changeHandlerTrue = function(err, status){
+    expect(status.code).toBe(13);
   };
 
-  var changeHandlerFalse = function(e){
-    expect(e.target.checked).toBeFalsy();
+  var changeHandlerFalse = function(err, status){
+    expect(status.code).toBe(12);
   };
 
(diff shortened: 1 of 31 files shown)


__tests__/Signature.js
```diff
@@ -5,34 +5,34 @@ var ReactDOM = require('react-dom');
...
-    receipt: 'somehash'
+    receipt: 'dc9e5827abb678f54103e7b89435abf9b36648797ebb6516a52ab33ab4e46cee'
   });
+  code = code === 13 ? 12 : 13;
 });
 
 describe('irs report', function(){
-  var changeHandlerTrue = function(e){
-    expect(e.target.checked).toBeTruthy();
+  var changeHandlerTrue = function(err, status){
```
awolfe76: I like the way you did things in the IRS test, for [example](https://github.com/cfpb/hmda-platform-ui/pull/142/files#diff-b28d2400315fa2e86b0451fc245b96bbR56). Both the function names and declaring them inside the `it()`.
Maybe do that here too?

awolfe76:
**There's a conflict: "This branch has conflicts that must be resolved".**

_The rest of this may be coming but wanted to point it out._

Still digging through it but when looking at the UI:
- for `Bank 0` after selecting a file and clicking upload no progress happens
- for `Bank 3`, it's "in progress" but when viewing the progress you see the file progress completes but no edits appear
- for `Bank 5`, its "complete" but there is no receipt and timestamp rendered (the checkbox is checked)
- **You mention is above** but I don't see the Q029 or Q595 edits on any of the banks
- Verification of a quality or macro edit removes the IRS report and signature

> allows transitioning to IRS report after verifying all quality/macro

I can't test this in the UI because several don't render the edits. _Seems_ to only render correctly when `code > 9`. I don't see edits unless the IRS report is there already.

wpears:
Currently ignoring q029/q595 wrt app state/flow.. since I think they should be rolled in with other quality edits 😎  (see #144)

awolfe76:
Looks like the latest commits fixed all but 

> Verification of a quality or macro edit removes the IRS report and signature

After entering text for a macro edit, clicking verify removes the IRS and signature components. The others are 👍  though.

wpears:
> Verification of a quality or macro edit removes the IRS report and signature

Which bank are you testing on? This is somewhat hacky because, in truth, IRS report and signature would never show if macro/quality haven't been verified

I'm guessing you're seeing this behavior because the mock data is in an impossible state.

That said, _unchecking_ a verified macro/quality _should_ remove IRS and signature components, no? Because they only show when all macro/quality are verified.

awolfe76:
I'm using `/bank6id/2017/2`.

You are right about 

> That said, unchecking a verified macro/quality should remove IRS and signature components, no? Because they only show when all macro/quality are verified.

But this is happening when I check a verification, which should do the opposite ... if all others are checked to. But it shouldn't remove those components.

wpears:
Right; but there is no ways those (IRS and signature) should be shown if everything isn't already validated. They only are now because we're using one set of edits for all the institutions. I can add additional json documents and the logic to route to the correct one, if you'd prefer. That'd probably be in a different PR, though

awolfe76:
Got it. I see what's going on now. Sorry for the confusion. Merging.

awolfe76:
And I think we're ok, no need to add more json. I think we'll get better once the back-end API stuff gets going.

PR #150, opened 2016-06-17
Edit Metadata - first pass

wpears:
Based off an internal document (Edit Requirements 2017), this is generated from an incredibly hacky awk script. The "name" is bungled in some cases ie, `totalNumberofLoanApplications` or `hOEPA`, both of which probably aren't what we'd want to use. Rather than finesse it in to shape with some dazzling regexp, I figured we should talk about what we want it to look like.

[Probably relevant doc from hmda-platform](https://github.com/cfpb/hmda-platform/blob/master/Documents/2017_File_Spec_LAR.csv)

### DIFF
--- a/edits.json
+++ b/edits.json
@@ -1,0 +1,744 @@
+{
+  "S010": {
+    "description": "The first row of a submission must be the transmittal sheet record and any subsequent row the Loan/Application Registers (LARs).",
+    "fields": {
+      "friendly": "Record Identifier",
+      "name": "recordIdentifier"
+    }
+  },
+  "S011": {
+    "description": "A submission must contain a transmittal sheet and at least one loan/application register (LAR)",
+    "fields": {
+      "friendly": "Record Identifier",
+      "name": "recordIdentifier"
+    }
+  },
+  "S020A": {
+    "description": "Agency Code must be populated with the values 1, 2, 3, 5, 7 or 9.",
+    "fields": {
+      "friendly": "Agency Code",
+      "name": "agencyCode"
+    }
+  },
+  "S020B": {
+    "description": "The Agency Code provided on both the Transmittal Sheet and the Loan/Application Record must be the same and they must also be the same Agency Code on record with the National Information Center.",
+    "fields": {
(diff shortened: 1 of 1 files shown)


awolfe76:
From chat the other day, one purpose of this file is to help in rendering additional data (description, field name, etc) about an edit:
addedcols

lesgou:
@wpears I think the format looks good. Some of the content is in draft and future state though, I can provide you with the updated.

debseidner:
@awolfe76  Like the picture. Would value entered contain more than one field's value if the edit relates to more than one field?

wpears:
@awolfe76 I think it'll be more like

```
_____________________________
|             |             |
|   Loan ID   | Agency Code |
|_____________|_____________|
|             |             |
|     256     |      5      |
|_____________|_____________|

```

The `friendly` name goes in the header and the value submitted is what fills the column

awolfe76:
Yep @wpears that's it. Thought something felt a bit off on mine.
@debseidner - what Wyatt has is correct. Each field gets a column.

awolfe76:
From our conversation the other day, it sounded like this data could potentially come from the backend. That the API response could contain this information, along with the values for each field required in the edit. Changing the API response to something like (see [this json file](https://github.com/cfpb/hmda-platform-ui/blob/master/server/json/syntactical.json) currently used by the front-end as a reference):

_The json is just an example. The point is that the field names and the values entered would be part of the response._

```
{
  "type": "syntactical",
  "edits": [
    {
      "edit": "S020A",
      "description": "Agency Code must be populated with the values 1, 2, 3, 5, 7 or 9.",
      "fields": [
        {
          "friendly": "Agency Code",
          "name": "agencyCode"
        }
      ],
      "lars": [
        {
          "lar": {
            "loanId": "s1",
            "values": [
                {
                    "value": "4",
                    "name": "agencyCode"
                }
            ]
          }
        }
      ]
    }
  ]
}

```

Is that right?

lesgou:
@wpears @awolfe76 The friendly language has been updated and saved where it was before.

PR #153, opened 2016-06-22
Remove text area for Quality edits

### LINKED ISSUE #145: Remove verification textareas from quality edits
It appears that quality edits do not need individual text areas for justification, they are just verified or not with a checkbox. If we feel this is the right step to take, we'll need to strip them out of the appropriate edit component.

awolfe76:
Closes #145 
- simplify rendering by modifying json (more to come out of backend API work)
- renders quality with only the checkbox
  - leaving text area for macro, along with the checkbox


awolfe76:
Closing. #159 and #160 will handle this.

PR #156, opened 2016-06-23
Verify all

awolfe76:
**Needs #153 merged first**

Closes #146 
- adds in the **single** verify all for each quality check
- passes the `type` of edit as a prop so that it can render only for quality edits
- currently no interaction (no API calls) on the checkbox
  - will wait for https://github.com/cfpb/hmda-platform/pull/422

Also created #157 to simplify some of this.

Issue #155, opened 2016-06-22
Branding


awolfe76:
We're going to be branding the UI as FFIEC. We can use this https://standards.usa.gov/ to help.
- uses Sass instead of Less
- different "style guide", USDS instead of CFPB.

wpears:
I have some reservations about this; will fill in more after a discussion on 7/6.

wpears:
Ignore the above; full speed ahead

awolfe76:
Breaking this into a few chunks (a few PRs):
- [ ] general clean-up to get ready
  - remove less and add sass, remove cf- dependencies, add uswds dependency, setup build
- [ ] begin to apply new styles

PR #159, opened 2016-06-28
Demo updates

wpears:
This PR proposes merging the changes made for the demo into master

This includes removing text areas for quality edits (probably should have PR'd this before OSBridge ;D), adding a "No errors" message when no edits fail, and a general info component that explains the needed response based on the edits found (refile, verify, etc.)

Also, refactors the status update callback to actually be an object with get and set methods, so child components can react to parent state in a reasonable way.

Also, changes the dev server to allow a progression between app states based on submission count: submission 1 will always find syn/val/qual/macro edits, submission 2 will always find qual/macro edits, submissions 3+ will find no edits.

The code in this PR is fully redundant with #153 and #156 will likely need some work to merge nicely with this. I'll probably just spend the evening reimplementing #156 on top of this PR 💃


src/js/EditsContainer.jsx
```diff
@@ -23,7 +23,7 @@ var EditsContainer = React.createClass({
   },
 
   propTypes: function(){
-    setAppStatus: React.PropTypes.func
+    appStatus: React.PropTypes.object
```
awolfe76: Could we also define that the object should contain functions?
https://facebook.github.io/react/docs/reusable-components.html

``` javascript
appStatus: React.PropTypes.objectOf(React.PropTypes.func)
```

_Probably in all cases below._

src/js/EditsDetail.jsx
```diff
@@ -5,7 +5,7 @@ var EditsDetail = React.createClass({
   propTypes: {
     details: React.PropTypes.array,
     primary: React.PropTypes.string,
-    setAppStatus: React.PropTypes.func
+    appStatus: React.PropTypes.object
```
awolfe76: Same as above ...
Could we also define that the object should contain functions?
https://facebook.github.io/react/docs/reusable-components.html

src/js/InstitutionStatus.jsx
```diff
@@ -28,7 +28,7 @@ var InstitutionStatus = React.createClass({
     var submission = this.props.institution.currentSubmission;
     var submissionRoute = submission === 0 ? 1 : submission;
     var appLink = '/' + id + '/' + period + '/' + submissionRoute;
-    var statusLink = <Link to={appLink}>View filing status</Link>
+    var statusLink = <Link className="statusLink" to={appLink}>View filing status</Link>
```
awolfe76, line 31: I don't think this class is necessary. I like the styling but I think its easier to solve it in other ways, e.g. wrapping it in a simple `<p>`.
_Same for others below._

src/js/RefileWarning.jsx
```diff
@@ -0,0 +1,39 @@
...
+
+var refileText = 'Syntactical and validity edits require file resubmission.'
+var validateText = 'Quality and macro edits must be validated before continuing.'
+
+var RefileWarning = React.createClass({
+
+  getText: function(){
+    var textToRender = null;
+
+    if(this.props.code === 7){
```
awolfe76, line 17: Because of `this.props.code` we should have a `propTypes: {}`

src/js/Signature.jsx
```diff
@@ -3,7 +3,7 @@ var api = require('./api');
 
 var Signature = React.createClass({
   propTypes: {
-    setAppStatus: React.PropTypes.func.isRequired,
+    appStatus: React.PropTypes.object.isRequired,
```
awolfe76, line 6: See comment earlier about using `objectOf` but also, is there a reason that this has `isRequired` and the others don't?

src/js/EditsDetailRow.jsx
```diff
@@ -7,13 +7,18 @@ var EditsDetailRow = React.createClass({
     detail: React.PropTypes.object.isRequired,
     primary: React.PropTypes.string,
     id: React.PropTypes.number,
-    setAppStatus: React.PropTypes.func
+    appStatus: React.PropTypes.object
+  },
+
+  defaultProps: {
```
awolfe76: Is there a reason to set these here and not in other places?

src/js/HomeLink.jsx
```diff
@@ -4,7 +4,10 @@ var Link = require('react-router').Link;
 var HomeLink = React.createClass({
 
   render: function(){
-    return <Link className="HomeLink" to='/'>Home</Link>
+    return <div class="AppContainer">
```
awolfe76, line 7: I _think_ I originally did this and now it looks odd. No worries on this PR though, I'll fix when we move to USDS in #155

src/less/components/InstitutionStatus.less
```diff
@@ -6,11 +6,21 @@
 
   h3 {
     font-family: "AvenirNextLTW01-Medium";
-    margin-bottom: 0;
+    margin-bottom: 0.234em;
```
awolfe76, line 9: `0.234em`?

src/less/components/InstitutionStatus.less
```diff
@@ -6,11 +6,21 @@
...
   }
   h5 {
     color: @gray;
     margin-bottom: 10px;
     margin-top: 0;
   }
 }
+
+.statusLink {
+  margin-bottom: 0.9375em;
```
awolfe76, line 19: I did suggest that this class isn't necessary, but `0.9375em`?

awolfe76:
We need tests for `RefileWarning.jsx`.

And since this PR is already fairly large, we should add tests for `SubmissionContainer.jsx` as well. It might be the most important component we have. 😄

src/js/EditsContainer.jsx
wpears: Sure! Good idea

src/js/RefileWarning.jsx
wpears, line 17: 👍

src/less/components/InstitutionStatus.less
wpears, line 19: It appears in capital framework as a margin (I chose it via inspecting styles).

src/less/components/InstitutionStatus.less
wpears, line 9: 1/4 of 0.9375 ^^;

src/js/InstitutionStatus.jsx
wpears, line 31: I can see what you mean here, but imo, the class is a good choice here. I'd say since the objective is purely stylistic (more whitespace here), using a class with styling information is truer to the intent than wrapping the element.

awolfe76, line 31: Yeah, I get it. I'm just more in favor of using pre-existing elements and styles rather than creating the one-offs. Something really clean about taking advantage of simple HTML.

A couple possible options:
1. **my recommendation**, ignore any CSS comments because we're going to go with the USDS stuff in #155 and I'll likely do some re-writing anyway
2. if you want ignore option one 😄 , rename the class to something more re-usable ... this is something I've been thinking about across the project
   - the component based CSS has advantages in some cases, but limits re-use in most (its why i created #116)
   - i think we need to do this only in very specific cases, more to come in #155

wpears:
@awolfe76 Think this captured everything; ticketed #162 to do separately as I want that to be very extensive and would like the current PRs to be merged first.

awolfe76:
Looks good to me 👍 
I totally agree with #162, its the keeper of all the things.

PR #160, opened 2016-06-28
Demo verify all

wpears:
Redoes #156 on top of #159, while also implementing the mocked backend logic for "verified" state updates and state transitions in the app.

Depends on #159 

Needs tests before merge; but feel free to give it a 👀


src/js/ValidationProgress.jsx
```diff
@@ -49,7 +49,7 @@ class ValidationProgress extends React.Component {
   }
 }
 
-ValidationProgress.propTypes = {initialCode: React.PropTypes.number};
-ValidationProgress.defaultProps = {initialCode: 3, appStatus: {get: function(){}, set: function(){}}};
+ValidationProgress.propTypes = {initialCode: React.PropTypes.number, appStatus: React.PropTypes.objectOf(React.PropTypes.func).isRequired};
```
awolfe76: Minor but this would be easier to read if broken down:

``` javascript
{
  initialCode: React.PropTypes.number,
  appStatus: React.PropTypes.objectOf(React.PropTypes.func).isRequired
}
```

wpears:
@awolfe76 should be good to go ^^;

PR #170, opened 2016-07-13
Macro select

wpears:
Changes over the EditsMacro component to use preset selections for justification, implemented as checkboxes because it is the most obvious way (imo) to select multiple items from a small list.

Also allows the specification of an "Other" selection which spawns a textarea that can be used to justify results with freeform text.


awolfe76:
👍  but thought maybe others wanted to take a look. If you don't hear anything feel free to merge.

PR #171, opened 2016-07-14
Nginx docker

wpears:
- Based on #170 
- Switches docker setup to nginx
- Includes dockerized dev server via `api-dockerfile`
- Includes a dev `docker-compose.yml` that runs the current static and api mock setup, but via docker


wpears:
Should be good to go; testing expose stuff completed

schbetsy:
LGTM. We also need updated instructions in the README for running the app in dev. I'll add an issue for that.

PR #174, opened 2016-07-27
Update nginx and docker-compose for running with back-end

schbetsy:
**Not ready for merge**

Issue #163, opened 2016-06-28
Implement Redux


awolfe76:
We're getting close to this, from the [motivation for Redux](http://redux.js.org/docs/introduction/Motivation.html):

> Managing this ever-changing state is hard. If a model can update another model, then a view can update a model, which updates another model, and this, in turn, might cause another view to update. At some point, you no longer understand what happens in your app as you have lost control over the when, why, and how of its state. When a system is opaque and non-deterministic, it’s hard to reproduce bugs or add new features.

[Redux](http://redux.js.org/index.html) will help.

awolfe76:
From standup discussion on 7/29:

The first round of Redux implementation will be for the home page only, to display the list of institutions. The PR to resolve this will likely resolve #176 as well.

This will also allow us to deploy a functioning integration to the cluster.

awolfe76:
Closing, #180 handles this.

PR #175, opened 2016-07-28
Update nginx setup and compose to work with the backend

wpears:
Takes the concepts in #174 and runs wild
- Sendfile and tcp stuff is static file performance stuff
- Max body size is set to 1GB so large filers won't have issues
- try files -> 404 on files with expected extensions and try files -> default to index.html so that the single-page app is served when a user refreshes after routing with client side browser history stuff
- Hours of debugging a "weird" issue with uploading large files, figured out that nginx will become sad if you send headers too early from an upstream server... [After receiving and interpreting a request message, a server responds
  with an HTTP response message.](https://www.ietf.org/rfc/rfc2616.txt). In retrospect this seems kinda obvious, but I had this beautiful idea in mind that the client would get a response, say a 202, _riiiight away_, so in the mock server I was just sending a response immediately in the route.. This was not the right thing.
  - Also trimmed font junk to strip away pesky errors

### DIFF
--- a/src/js/actions/index.js
+++ b/src/js/actions/index.js
@@ -1,0 +1,1 @@
+//Actions here
--- a/src/js/actions/index.js
+++ b/src/js/actions/index.js
@@ -1,1 +1,0 @@
-//Actions here
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,11 +1,11 @@
 # dev setup assumes a mock-api image is built from api-dockerfile
 version: '2'
 services:
-  hmda-api:
+  api:
     image: mock-api
   ui:
     build: .
     ports:
       - "80:80"
     depends_on:
-      - hmda-api 
+      - api 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,9 +2,13 @@
 version: '2'
 services:
   api:
+    volumes:
+      - ./server:/usr/src/app/server
     image: mock-api
   ui:
     build: .
+    volumes:
+      - ./dist:/usr/src/app/dist
     ports:
       - "80:80"
     depends_on:
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -16,6 +16,8 @@ http {
   server {
     listen 80;
     access_log /dev/stdout;
+    gzip on;
+    gzip_types application/javascript text/html text/css application/json;
 
     # Whitelisted extensions
     location ~ \.(html|css|js|png|jpg|svg|eot|ttf|woff)$ {
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -10,23 +10,36 @@ http {
   include  mime.types;
 
   upstream api{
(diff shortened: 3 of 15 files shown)


src/js/actions/index.js
```diff
@@ -0,0 +1 @@
+//Actions here
```
awolfe76: This file looks like it might be from work on #163, probably don't need it here?

awolfe76:
I'm not exactly sure when we will use the `docker-compose.yml` file? The same for the [`api-dockerfile`](https://github.com/cfpb/hmda-platform-ui/blob/master/api-dockerfile) from the [earlier PR](https://github.com/cfpb/hmda-platform-ui/pull/171/files).

wpears:
The docker-compose file in this repo will be for when we want to run the app off the mock API rather than the backend, which will still be useful for some time, until all the backend endpoints are completed.

There's something to be said about having the back-end mock up unimplemented endpoints instead of the front-end, but if/until that happens this docker-compose is a nice way to get the project running in separate containers with returns for all the needed endpoints.

awolfe76:
Yeah, the leaving the mock around for a bit makes sense until the backend is up and running.

But can't we just continue to use the mock as we have been? Is there a benefit to us running that in docker? Seems like its going pretty smooth so far.

wpears:
That's true, but I think this will make it easier for the whole team to be able to run the mock api and also keeps docker front-of-mind since ultimately backend integration is our goal.

awolfe76:
I'm fine adding it. Just worry a bit about adding files the are likely to removed in near future. We'll clean it up though.

schbetsy:
When I run `docker-compose` from the **back end**, I see 
- At `/`, the homepage with the FFIEC logo and nothing else. In the console, there's the error we saw when we paired: "Unexpected institution status"
- At `/api/institutions` and other completed api endpoints, I see valid json responses from the back-end.

When I run `docker-compose` from the **front end** with the mock api, I see
- At `/`, the full homepage with Jane Doe's 6 banks and their info
- At `/api/` (and `/api`? It's hard to tell), there's a console error: `net::ERR_TOO_MANY_REDIRECTS`
- At `/api/institutions`, there's mock json data for institutions.

@wpears Does this match what you're expecting?

wpears:
@schbetsy `/api/` and `/api` with the mock api should now show `Welcome to the api`
https://github.com/cfpb/hmda-platform-ui/issues/176 created to sync what the front-end and backend return at /institutions and make this what the front-end expects.

awolfe76:
@wpears - while we're at it, want to enable `gzip` in `ngnix.conf`?

schbetsy:
LGTM

awolfe76:
Also, since the `docker-compose.yml` is meant to use for local dev, want to add in a `volumes:`? That way we can still use the `watch` task. (Same as described in https://github.com/cfpb/hmda-platform/pull/460#issuecomment-235938168.)

wpears:
Should be g2g

Issue #180, opened 2016-08-01
Continue implementation of Redux and ES6


awolfe76:
# This issue will be used to track the implementation of Redux across the components. While doing the implementation we will also be updating the components to use ES6. (Closing #133 in favor of this issue.)

This issue will be closed over several pull requests, updating one or a few components at a time. Its possible that some components do not need updated by the Redux implementation, so at a minimum they all will be updated to use ES6.

**Update:** also as part of this issue we should add comment where necessary.

List of components to update:
- [x] AppContainer (covered in #163)
- [x] DivisionHeader (#206)
- [x] EditReports (#216)
- [x] EditsContainer (#216)
- [x] EditsDetail (#216)
- [x] EditsDetailRow (#216)
- [x] EditsGrouped (#216)
- [x] EditsHeaderDescription (#216)
- [x] EditsMacro (#216)
- [x] ~~EditsQ029 (wyatt)~~ moved to quality
- [x] EditsQ595 (#216)
- [x] EditsSelector (#216)
- [x] HomeLink
- [x] index (covered in #163)
- [x] InstitutionContainer (probably covered in #163)
- [x] InstitutionStatus (probably covered in #163)
- [x] IRSReport (#211)
- [x] Multicheck (#216)
- [x] Progress
- [x] RefileWarning (#215)
- [x] Resubmit
- [x] Signature (#211)
- [x] SubmissionContainer
- [x] Summary (#213)
- [x] UploadForm
- [x] UserHeading (#206)
- [x] ValidationProgress

PR #181, opened 2016-08-02
Institutions page - redux

wpears:
Will fail tests, beware. Putting up this PR as a place for comments/discussion on the redux implementation so it isn't just living in my branch.

Structure is fairly idiomatic redux, but please do ask questions if you have them.

The groupings on the homepage were omitted here for simplicity of implementation, but will be added back in short order.


awolfe76:
@wpears - is this one ready for full review? _(There are some conflicts too.)_

wpears:
Yep, ready to go; I have another not quite baked PR with a large percentage of the redux changes coming in next.

src/js/InstitutionContainer.jsx
```diff
@@ -7,6 +7,7 @@ var InstitutionStatus = require('./InstitutionStatus.jsx');
 var InstitutionContainer = React.createClass({
 
   getInitialState: function(){
+    console.log('getting initial');
```
awolfe76, line 10: Can you remove the logging? If you need it for the next PRs that's ok, this can just be a reminder then.

awolfe76, line 10: Actually, can this whole file be removed?

state_shape.json
```diff
@@ -0,0 +1,70 @@
+{
```
awolfe76, line 1: If this is really for documentation purposes, maybe a `.md` with some comments?

src/js/reducers/index.js
```diff
@@ -0,0 +1,39 @@
...
+}
+
+export const filings = (state = [], action) => {
+  switch (action.type) {
+  case RECEIVE_INSTITUTION:
+    return [
+      ...state,
+      ...action.institution.filings
+    ]
+  case CLEAR_FILINGS:
```
awolfe76, line 29: Maybe some comments around why we are returning the new state in these ways? Why its set to return what its returning? Might be helpful later to know why.

src/js/actions/index.js
```diff
@@ -0,0 +1,67 @@
...
+}
+
+export function clearFilings() {
+  return {
+    type: types.CLEAR_FILINGS
+  }
+}
+
+export function fetchInstitutions() {
+  return dispatch => {
```
awolfe76, line 39: I think some comments in some of these actions would be helpful.

awolfe76:
Overall 👍 . I think comments in some places, maybe more than I put line comments on, would be helpful. Had been thinking that for a while. Maybe as part of this redux/es6 refactor we should add some comments.

awolfe76:
@wpears - if you think you can get https://github.com/cfpb/hmda-platform-ui/issues/187 into this PR it would be good too. That should get us to a point where we can see the home page listing the institutions.

src/js/InstitutionContainer.jsx
wpears, line 10: This file is removed in the next PR

state_shape.json
wpears, line 1: Yeah, this needs to be updated in general, maybe a separate PR?

wpears:
I'll do #187 in a separate PR and comments/state shape updates in a yet another. Sound okay?

awolfe76:
I updated #180 to include adding comments.

PR #185, opened 2016-08-04
Source maps full

awolfe76:
- adds arg to `Dockerfile` and defaults to `build`
  - this is the production build
  - it is overwritten in the docker-compose setup for local development

Also see https://github.com/cfpb/hmda-platform/pull/482


kgudel:
Got it all working and it properly referenced the source files, for example assigning the error `throw new Error('Unexpected institution status');` to InstitutionStatus.jsx instead of app.min.js 

cfpb/hmda-platform#482

awolfe76:
@wpears - to finish this off (another PR) I was trying to simplify the process by simplifying it a bit and making the source maps for both Less/CSS and JS external files.

I've pushed a working branch up, `testing-source-maps`.

I managed to get the Less/CSS working, inside a container and using the npm build. See the `prod:less` script.

For the JS I'm close. But Chrome is telling me `Failed to parse SourceMap` when inside the container. Oddly enough, it does seem to work when just running `npm start`. I'm running the `js` script and then the `js:min` script. You can see that in the `build:docker` of that testing branch.

There are a few other packages that might be worth looking at, they claim to handle browserify with source map (-d) -> uglify with source map:
- https://www.npmjs.com/package/uglifyify
- https://www.npmjs.com/package/minifyify

The end goal of this last part is to make it easier. To make it so we only have one `build` command (removing the need for `build:docker`) which does all the work. It would build the js and css and minify/uglify them with source maps, along with minifying the html and images (html and images are working already). We should be able to get away with the one `build` if we can get external source maps working. Minified versions during development won't be an issue in that case. We could also get rid of the `prod:js` and `prod:less` scripts, having them just be `js` and `less`, since it will be the same no matter what environment we're building for.

If that all works, we would also be able to get rid of the `ARG` [here](https://github.com/cfpb/hmda-platform-ui/pull/185/files#diff-3254677a7917c6c01f55212f86c57fbfR4) and [here](https://github.com/cfpb/hmda-platform/pull/482/files#diff-4e5e90c6228fd48698d074241c2ba760R9). One build so no need for it.

Hope all that makes sense. If you get a chance would you mind taking a look.

_It's also possible that we don't need source maps in production, although it seems like it would be nice to be able to debug easily there as well._

wpears:
Solved by #190

PR #186, opened 2016-08-16
More redux

wpears:
Has redux code and tests through the app flow up to but not including the various edit components (separate PR). Depends on #181


awolfe76:
@wpears - this PR is still showing commits from #181, making it hard to view. Want to work some magic and update it?

wpears:
Fixed

src/js/components/ValidationProgress.jsx
```diff
@@ -0,0 +1,38 @@
...
+const ValidationProgress = (props) => {
+
+  const code = props.status.code
+
+  let uploadComplete = null
+  let parsingStarted = null
+  let parsingComplete = null
+  let validationStarted = null
+  let validationComplete = null
+
```
awolfe76, line 12: Do you think we need a `if(code == 2) uploadingStarted = <p>Uploading ...</p>`?

And, I think the `parsingStarted` and `validationStarted` should be removed from the UI when the `Complete` state is reached. (Same for `uploadingStarted` if we add it.)

src/js/containers/Institutions.jsx
```diff
@@ -37,9 +37,4 @@ function mapStateToProps(state) {
   }
 }
 
-InstitutionContainer.propTypes = {
```
awolfe76, line 40: Why did you take this out?

src/js/containers/ValidationProgress.jsx
```diff
@@ -0,0 +1,20 @@
+import { connect } from 'react-redux'
+import ValidationProgress from '../components/ValidationProgress.jsx'
+
+function mapStateToProps(state) {
+  const {
+    submission
+  } = state.app.submission || {
+    subimission: {
```
awolfe76: Typo here: sub**i**mission.

src/js/containers/ValidationProgress.jsx
```diff
@@ -0,0 +1,20 @@
...
+      id: 1,
```
awolfe76, line 9: Not sure I follow why we would default this to 1?

src/js/containers/ValidationProgress.jsx
```diff
@@ -0,0 +1,20 @@
...
+      status: {
+        code: 3,
```
awolfe76, line 11: Same as above, why 3?

src/js/containers/Institutions.jsx
wpears, line 40: For a connected component it's kinda redundant, it would be checking react-redux and react-router rather than anything we've written

wpears, line 40: (params are passed automatically by react-router and dispatch by react-redux)

src/js/components/ValidationProgress.jsx
wpears, line 12: Validation progress is only rendered when the upload completes and there is a separate progress tracker for the upload.

Also, I'd be fine changing the design of this component but maybe in another PR (this PR purposely has minimal style)

src/js/containers/ValidationProgress.jsx
wpears, line 9: 1 here assuming that submissions are given increasing integers as ids. The default is only provided here to ensure default props are passed to the presentational component (of course that component could have its own default props, but I think that adds slightly too much smarts to the presentational component)

src/js/components/ValidationProgress.jsx
awolfe76, line 12: Created #192 for this.

src/js/containers/ValidationProgress.jsx
wpears, line 11: 3 because the validation progress component is first rendered when the upload is complete, meaning this is a semi-sane default.

PR #199, opened 2016-08-24
Edits redux

wpears:
Contains the barebones edits implementation in redux.

Still needs:
- tests
- justification checkboxes and associated actions

Feel free to peruse in the meantime


awolfe76:
👍  plenty more to go but good here.

I think we'd be ok for now to leave the expandables out completely. We know what to do but lets wait and see if its really necessary. There's some data to back this up, although not for syntax and validity. Plus, we may end up only showing syntax and validity first (based on todays conversation), so I think we're ok to hold off. Just wanted to say that so we don't include that work yet. Keep it simple first.

PR #201, opened 2016-08-24
update to the correct port for the backend

awolfe76:
quick update to the correct port for the backend api.


awolfe76:
Merging for quick fix.

Issue #202, opened 2016-08-25
Fix docker-compose


awolfe76:
https://github.com/cfpb/hmda-platform-ui/pull/201 was to get the cluster deployment working. But it broke the docker-compose setup.

For now, to run docker-compose locally, you have to manually update the https://github.com/cfpb/hmda-platform-ui/blob/968f5635a85274a218c08b34ee5247979e3c07a2/nginx/nginx.conf#L13 to use port `8080` instead of `10003`.

This issue may require changes to the [platform](https://github.com/cfpb/hmda-platform) repo too.

awolfe76:
Best answer here maybe to have back-end update configuration to use port `10003` instead of `8080`, leaving the `nginx.conf` as it is.

awolfe76:
@jmarin @hkeeler what do you think about updating the back-config and then doing `EXPOSE 10003` in the dockerfile? I could recreate this issue in the platform repo if so.

hkeeler:
@awolfe76 I agree.  Due to Nginx's static configuration, and the differences in how service discovery works in Compose vs our internal setup, the only _simple_ fix here is to make the container port and Docker host port be the same value (in this case `10003`) for any Nginx `upsteam`s.  And yes, the cleanest way to do this is to update [`hmda-platform/Dockerfile`](https://github.com/cfpb/hmda-platform/blob/master/Dockerfile#L18) to `EXPOSE 10003`.

Looks like we came to that same conclusion for grasshopper (see [`grasshopper/docker-compose-common.yml`](https://github.com/cfpb/grasshopper/blob/master/docker-compose-common.yml#L4))...which should also be updated to reflect the very similar change made to [`grasshopper-ui/nginx/nginx.conf`](https://github.com/cfpb/grasshopper-ui/blob/master/nginx/nginx.conf#L19).

@awolfe76, feel free to create those issues.  I think you can close this one out, as I don't think there should be any changes to this repo as part of this.

wpears:
Hah, perhaps I should have read this before pushing https://github.com/cfpb/hmda-platform-ui/pull/203, which I think is actually a cleaner approach as it separates the dev setup from deploy.

Imagine if/when the marathon service port changes, we'd have to update code both in this project and in hmda-platform to keep everything in sync. With the code in #203, you just update the envvar in Jenkins.

awolfe76:
I agree @wpears. Nice solution!

[...]

PR #203, opened 2016-08-25
Create HDMA_API_PORT scheme for seamlessness between dev, compose, and deploy

wpears:
If the envvar `HMDA_API_PORT` is set, that will be the upstream port, otherwise it will be 8080. This can be set on dev machines if desired and in jenkins on build

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,13 +11,14 @@ RUN apt-get update && \
 RUN mkdir -p /usr/src/app
 WORKDIR /usr/src/app
 COPY . /usr/src/app
+COPY nginx/nginx.conf /etc/nginx/nginx.conf
 
 RUN npm cache clean && \
     npm install && \
     npm run clean && \
-    npm run build
-
-COPY nginx/nginx.conf /etc/nginx/nginx.conf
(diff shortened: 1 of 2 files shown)


awolfe76:
Closes #202

awolfe76:
👍  from me. I see the successful build in jenkins and see the app in marathon. Also ran the local docker-compose set up and its good to go!

hkeeler:
I'm OKish with this.  It is certainly more flexible, and gets us away from having to change all our images whenever an internal port changes.

However, we should probably have a discussion about if/how we want to use build-time `ARG`s in general.  The Docker team said _no_ to this feature for a long time for good reason.  It breaks the general contract of being able to know everything about an image based on just the `Dockerfile`.  Now you need to know what args were passed in at build time too.

Why not do the `sed` find/replace at runtime before starting nginx?  That's the pattern I've generally seen for when a given service has a static config that needs to be updated based on a given environment.  Then the envvars get set in the standard docker runtime way, and we don't need to know how the image was built.  Is there something special about our setup that is forcing us into doing this at build time?

PR #206, opened 2016-08-30
ES6 updates

awolfe76:
Part of #180 

Updates to components and tests (ES6 only) for:
- DivisionHeader
- HomeLink
- UserHeading


awolfe76:
@wpears this is ready when you have time.

PR #208, opened 2016-09-02
Signature and IRS report redux

awolfe76:
@wpears I put this up as a PR because i thought it might be easier to see but its definitely a WIP.

I'm getting an error when rendering the IRS component:
`warning.js:44 Warning: There is an internal error in the React performance measurement code. Did not expect componentDidMount timer to start while render timer is still in progress for another instance.`
That error is pretty bad. 😦  Google isn't really helpful, at least to me.

You can see the signature component rendered if you comment out the `irs = <IRSReport checked={false} />` in `/containers/Submission.jsx` on [line 49](https://github.com/cfpb/hmda-platform-ui/compare/sign-irs-redux?expand=1#diff-393ea7c3c4a7c4a139361de56866fd01R49).

I think its pretty close. Would like to walk through it a bit on Tuesday when we're back.

Also, if #206 is merged this will be even easier to see. 😄


wpears:
So your weird error is just downstream noise from including the incomplete `<Edits/>` component. Remove that and you're good to go (you'll probably still need to debug things).

re: component design, I think you'll want to split these into containers and components; where the container handles the logic and component the bare presentation (shouldn't have any methods). Connect will be run from the container, whose `mapStateToProps` helps decides the child props. (side note: how you currently have `propTypes` is a bit wonky, because you're defining the proptypes on the child component that is then wrapped by connect, which is what is exported.)

Finally, I think it is very important that "checked" is not passed into these components from the `submission` component. Instead, it should be modeled as part of the state which then informs the props via `mapStateToProps`.

wpears:
(one of the things you'll notice when you exclude the edits component is that the irs component is referring to `self.props.irs.msas` when it is undefined. This has to do with the proper default props as said above)

wpears:
(one of the things you'll notice when you exclude the edits component is that the irs component is referring to `self.props.irs.msas` when it is undefined. This has to do with the proper default props as said above)

awolfe76:
👍  working through it now.

awolfe76:
@wpears - this is ready now. 

Notes:
- there is no `POST`, created #210
- tests failing, will fix in #210

wpears:
yolo breaking master

PR #209, opened 2016-09-02
Match institutionIds in header to demo data in platform

schbetsy:
This will be necessary when [PR #555](https://github.com/cfpb/hmda-platform/pull/555) on Platform gets merged.


awolfe76:
👍  makes sense to me. Thanks @schbetsy.

PR #211, opened 2016-09-16
Signature and IRS redux

### LINKED ISSUE #210: Update IRS and Signature to handle POST
#208 got these components most of the way, but we also need to be able to handle the `POST` for each when a user checks/un-checks the checkbox.

awolfe76:
Closes #210. Progress toward #180.
- implements redux on the signature and irs components
  - both GET and POST are handled
- updated the API and responses (see `server/json/irs.json` and `server/json/receipt.json`)
- also implements a new way to handle the status of the application


awolfe76:
@wpears - tests are passing!

src/js/api.js
```diff
@@ -54,27 +55,27 @@ export function getLatestSubmission(){
 }
 
 export function getEditsByType(submission){
-  return sendFetch(`/submissions/${submission}/edits`)
+  return sendFetch('/submissions/${submission}/edits')
```
wpears: These should be the original grave marks because we're using ES2015 string substitution here

wpears: (applies throughout the file)

wpears (changes requested):
Everything looks good, but the api grave stuff will need to be reverted before merge

src/js/api.js
awolfe76: 👍  - ready to go.

Issue #205, opened 2016-08-26
Don't post to /latest for submissions


awolfe76:
We should always post files to `/<submissionID>`. And possibly `/submissions`.

See https://github.com/cfpb/hmda-platform/issues/535 for the discussion.

hkeeler:
We've also discussed `POST`ing the file to `institutions/{institution_id}/filings/{period}/submissions/` directly, and having that action be what creates a new `submission_id`.  This flow makes more sense to me.  After all, what is a submission without the file?

This would obviously be a change to both frontend and backend.  This could be spun off to its own issue as well if this one is intended as a quick fix.

awolfe76:
I do have that listed on [this issue](https://github.com/cfpb/hmda-platform-ui/issues/204) (integration tracking) and [this one](https://github.com/cfpb/hmda-platform/issues/537) (endpoint creation tracking). And it makes sense to me too.

Updated the comment above with it.

PR #215, opened 2016-09-21
Refile Warning

### LINKED ISSUE #214: Redux - RefileWarning component
Implement redux and ES6 on the RefileWarning component.
- [x] ES6
- [x] update `/container/Submission.jsx` to render the refile warning
- [x] tests

awolfe76:
Closes #214.
- updates refilewarning component to ES6 and redux
- adds in the container
- updates tests

### DIFF
--- a/__tests__/RefileWarning.js
+++ b/__tests__/RefileWarning.js
@@ -1,38 +1,62 @@
-jest.dontMock('../src/js/RefileWarning.jsx');
+jest.unmock('../src/js/components/RefileWarning.jsx');
 
-var React = require('react');
-var ReactDOM = require('react-dom');
-var TestUtils = require('react-addons-test-utils');
+import RefileWarning from '../src/js/components/RefileWarning.jsx'
+import Wrapper from './Wrapper.js'
+import React from 'react'
+import ReactDOM from 'react-dom'
+import TestUtils from 'react-addons-test-utils'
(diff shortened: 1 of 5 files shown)


wpears:
I think we'll need to make a decision about how new submissions before this can be merged; you'll notice that the "Refile here" link is busted. This is because the assumptions around how submissions are created is somewhat nebulous. Possible options:
- Just (clear the component) and reupload a new file. The act of reuploading triggers the backend to create a new submission under which this upload (and subsequent actions) are categorized. We're doing this implicitly now by using `/submissions/latest` for the upload path. In this case we'd probably either want to just drop the "Refile here" link or make it empty the state so that the edits go away and the upload component goes to its default state
- Submissions are created explicitly via an API call. This means we'd have to call this in the Upload action before actually sending the file over (and we'd change the upload URL to point to `submissions/<sub id>`). The "Refile here" link would also hit this endpoint and cause the errors to clear automatically (new submission with new state). This is probably more obvious to the user that they are starting over.

wpears:
See https://github.com/cfpb/hmda-platform-ui/issues/205 and https://github.com/cfpb/hmda-platform/issues/535.

Probably most obvious/easiest to just have an endpoint to create a new submission (option two above); but either option could work.

wpears:
Closed by #228

PR #216, opened 2016-09-22
Edits/Justification

wpears:
Uses react-select for the justification drop down; pretty nifty

Kinda weird integration with redux (since this react component manages its own state in a non-redux-y way), but it still triggers actions/reducers so that the redux model is of the state is always in sync with react-selects internal model.

Tests pass but coverage is down (more tests needed)

Need sort functionality, and to decide how this impacts modeling of the edit state... (modeling of the state needs to be looked at a little closer anyhow, since we'll need a paging solution)

For now though, :eyes: check it out 👀


server/routes/edits.js
```diff
@@ -5,14 +5,12 @@ var syntactical = JSON.parse(fs.readFileSync('./server/json/syntactical.json'));
 var validity = JSON.parse(fs.readFileSync('./server/json/validity.json'));
 var quality = JSON.parse(fs.readFileSync('./server/json/quality.json'));
 var macro = JSON.parse(fs.readFileSync('./server/json/macro.json'));
-var q595 = JSON.parse(fs.readFileSync('./server/json/q595.json'));
 
 var edits = {
     syntactical: syntactical,
     validity: validity,
     quality: quality,
     macro: macro,
```
awolfe76: Not a huge deal, but could probably remove the trailing comma.

awolfe76:
Still reviewing, but when running it and trying to add/remove justifications for macro edits nothing happens and I see a console error: `Uncaught TypeError: n.split is not a function`

This is coming from `JustificationSelector.jsx` line 87 -> `var values = value.split(delimiter);`

awolfe76:
I think I get why the state management within the component, but can you explain it a little bit? Might be helpful to have it here.

wpears:
react-select was built in such a way (afaik) that setting the component's state is how things happen. (things happen here being add to selected input & remove from selection list or the inverse).

Components aren't strictly denied being stateful in redux, just all state must flow from (or be captured in) the top-level store. With the implementation here, the standard api of react-select doesn't have to be hacked around and yet the actions that trigger UI updates are still captured and flow through reducers as necessary.

awolfe76:
@wpears - i pulled down your latest 2 commits, ran `npm install` and `npm build` again but am still seeing the same error. Am I missing a step?

wpears:
I'd bet you're installing from shrinkwrap; nuked it.

awolfe76:
That did it. Thanks!

Looks good to me 👍 . Merging.

PR #217, opened 2016-09-28
Update states

awolfe76:
Also see https://github.com/cfpb/hmda-platform/pull/589 which proposes a small update to the order of the states. That update doesn't change this PR.
- updates states to match back-end [states](https://github.com/cfpb/hmda-platform/blob/master/Documents/submission-status.md)
- removes documentation from here, we can use the [doc from the back-end repo](https://github.com/cfpb/hmda-platform/blob/master/Documents/submission-status.md)


awolfe76:
Good catch @wpears.

PR #219, opened 2016-09-30
Branding - Round 2

### LINKED ISSUE #155: Branding

awolfe76:
Depends on #218 

Closes #155 - **I'll start to make new issues to fix specific components.** For example, the IRS table needs some work, the institutions list as well, and plenty of other things 😄 . But this is a good start.
- adds in [uswds](https://standards.usa.gov/) imports
- grid layout updates
- styling for all components is close
- progress component renders as nav for use in the future (anchor links) and overwrites text as it, well, progresses
- uses `<caption>` as a secondary table heading, see https://www.w3.org/WAI/tutorials/tables/two-headers/
- other minor styling updates (with plenty more to come)

---

A few screenshots ...
#### Institutions list

uswds-institutions
#### Submission screen

uswds-submission


wpears (approved):
Cool. Looks like a good first pass. I think ticketing out smaller components (as you said above), is a winning strategy after the changeover

PR #228, opened 2016-10-04
Separate submission creation

wpears:
Also contains code for #214. Separates the uploading process, as far as the API is concerned, into an upload and a submission creation phase, which has some nice properties.


src/js/components/RefileWarning.jsx
```diff
@@ -0,0 +1,42 @@
...
+     }else{
+       textToRender = validateText
+     }
+
+  return <h3><span className="cf-icon cf-icon-error cf-icon__3x"></span><span className="refile-text">{textToRender}{refileLink?' ':''}{refileLink}</span></h3>
+}
+
+const getRefileLink = (props) => {
+  const location = parseLocation()
+  const href = `/${location.id}/${location.filing}`
```
awolfe76: Couldn't we just use `props.params` here? Then we could remove the `import { parseLocation } from '../api'` and not use `parseLocation()`. 

It could just be `return <Link to={'/${props.params.institution}/${props.params.filing}'} onClick={props.refileLink}>Refile here.</Link>`

awolfe76 (commented):
I'm also seeing `api.js:80 Uncaught Error: Must provide a submission id when data is uploaded.` when trying to upload a new file.

src/js/components/RefileWarning.jsx
wpears: This part is actually unnecessary, since it doesn't actually route you anywhere..! I just removed the href all together to simplify the component

wpears:
Doh. Commented out some vital code whilst bug fixing and didn't uncomment..! Yet another reason to keep improving the test suite.

awolfe76:
@wpears looks good. But do you think we should clear the file and the bytes uploaded bit when the refile link is clicked?

awolfe76:
Merging. Created #237 to handle the question above.

PR #253, opened 2016-10-21
render 'transmittal sheet' if ts:true

### LINKED ISSUE #250: Add transmittal sheet errors to the UI
Based on https://github.com/cfpb/hmda-platform/pull/623 we should now be able to output the transmittal sheet errors along with the other edits.

Each edit will have a `"ts": true,` set to true or false. If true, we can render "Transmittal sheet" as the "loan id".

We may need to have a different header for that column if it causes confusion.

### LINKED ISSUE #259: Fix file upload
Because of https://github.com/cfpb/hmda-platform/pull/655 we're getting a `406 Not Acceptable - Resource representation is only available with these types: application/json`.

Should be an update to https://github.com/cfpb/hmda-platform-ui/blob/1f87c8885fc2465164de38a8d6a7b2fb428a0c0f/src/js/actions/index.js#L332.

awolfe76:
Closes #250 and Closes #259 

- renders a row with "Transmittal Sheet" instead of loan id if `"ts": true`
- pass `type` to `renderTables` in `EditsByType`, we don't have `type` as a property of within a given edit type in the response
- accepts JSON response from file upload

### DIFF
--- a/__tests__/EditsByType.js
+++ b/__tests__/EditsByType.js
@@ -15,7 +15,7 @@ const types = {
   syntactical: JSON.parse(fs.readFileSync('./server/json/syntactical.json')),
   validity: JSON.parse(fs.readFileSync('./server/json/validity.json')),
   quality: JSON.parse(fs.readFileSync('./server/json/quality.json')),
-  macro: JSON.parse(fs.readFileSync('./server/json/macro.json'))
+  macro: {"edits": []}
 }
 
 const typesNoMacro = {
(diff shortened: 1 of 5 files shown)


awolfe76:
@wpears - i think this is ready now.

awolfe76:
@wpears - this PR will fix the file upload which is broken in the cluster right now.

awolfe76:
@wpears - not sure why Travis is failing. The tests pass locally.

wpears (approved):

Issue #244, opened 2016-10-17
Integrate with hmda-platform-auth


wpears:
This will involve changing the structure of some API calls as headers required by the backend will be added by the proxy based on authorization. As the mock api is being shelved in #243, there isn't too much meta work here outside of adjusting the calls and associated tests.

awolfe76:
@wpears - good to go here?

Issue #227, opened 2016-10-03
Which OpenID Connect JS lib should we use? 


hkeeler:
We need to decide which JS lib works best for us.  There aren't a ton of options out there for pure browser-based JS.
1. [Keycloak Javascript Adapter](https://keycloak.gitbooks.io/securing-client-applications-guide/content/topics/oidc/javascript-adapter.html)
   - The _Official_ Keycloak JS clients.
   - Definitely Keycloak-specific, not a general-purpose OIDC client.
   - Does have some nice features, like wrappers around API calls, so you don't make calls in the first place if your token has expired. 
2. [oidc-client-js](https://github.com/IdentityModel/oidc-client-js)
   - This project looks pretty solid.  It was recently [completely rewritten](https://brockallen.com/2016/05/25/oidc-client-re-released/), and sounds like it using the latest JS hotness.
   - Handles token storage and a bunch of other goodness with its [`UserManager`](https://github.com/IdentityModel/oidc-client-js/wiki#usermanager) object.
   - [Events](https://github.com/IdentityModel/oidc-client-js/wiki#events) fire when token is close to expiring, and other lifecycle events.
3. _Roll our own_...
   - There are lots of resources on this.  I based my original on [openidconnect.js](https://www.sakimura.org/test/openidconnect.js), which comes from [_Making a Javascript OpenID Connect Client in 4 steps_](https://nat.sakimura.org/2014/12/10/making-a-javascript-openid-connect-client/).  It's good for learning how OIDC works, but it is not production quality, and not maintained by anyone.
   - If we go down this route, I will dig up some more resources from when I originally started looking into this several months ago.

hkeeler:
Per standup today, we're going to give oidc-client-js a try.  I think we can use just the [basic config](https://github.com/IdentityModel/oidc-client-js/wiki#configuration), which would look like:
- **`authority`:** https://192.168.99.100:8443/auth/realms/hmda/.well-known/openid-configuration
  - This is the current URI.  The IP is obviously the Docker host IP, so it is subject to change based on your local setup.
- **`client_id`:** hmda-api
- **`redirect_uri`:** This is a URI we'll be adding to for the OIDC Provider's callback.
  - It will need to handle both [success](http://openid.net/specs/openid-connect-core-1_0.html#ImplicitAuthResponse) and [failure](http://openid.net/specs/openid-connect-core-1_0.html#AuthError) states of the OIDC Implicit Flow.
[...]

PR #257, opened 2016-11-02
Auth integration

wpears:
First pass at auth integration, depends on https://github.com/cfpb/hmda-platform-auth/pull/19
Know issues:
refreshing a page upsets the flow, ticketed in #258 
uploading a file results in a 400 ticketed in https://github.com/cfpb/hmda-platform-auth/issues/18


src/js/UserManager.js
```diff
@@ -0,0 +1,18 @@
+import { createUserManager } from 'redux-oidc'
+
+const userManager = createUserManager({
+  authority: 'https://192.168.99.100:8443/auth/realms/hmda',
```
hkeeler, line 4: We discussed setting these values at container startup, likely via some awk/find/replace goodness.  Do you want to do that as part of this PR, or a follow up?  As it is now, this will obviously only work in our local dev setups.

src/js/UserManager.js
```diff
@@ -0,0 +1,18 @@
...
+  client_id: 'hmda-api',
+  redirect_uri: 'http://192.168.99.100/oidc-callback',
+  silent_redirect_uri: 'http://192.168.99.100/silent_renew.html',
+  automaticSilentRenew: true,
+  scope: 'openid profile email',
```
hkeeler: You can remove `email`.  Keycloak doesn't honor that scope.

src/js/UserManager.js
```diff
@@ -0,0 +1,18 @@
...
+  response_type: 'id_token token'
+})
+
+//Mock client-side validation until Keycloak implements CORS on certs endpoint
```
hkeeler: Can this be removed now?

src/js/api.js
```diff
@@ -1,26 +1,36 @@
...
+  accessToken = token
+}
+
+export function getAccessToken() {
+  return accessToken || ''
 }
 
 export function makeUrl(obj, suffix){
-  var url = location.protocol + '//' + location.host + '/api'
+  var url = process.env.HMDA_API || location.protocol + '//' + location.host + '/hmda'
```
hkeeler: When do the `/hmda`-based URLs come into play?

src/js/containers/Login.jsx
```diff
@@ -0,0 +1,52 @@
...
+
+
+class LoginContainer extends Component {
+  constructor(props) {
+      super(props)
+  }
+
+  componentDidMount() {
+    const manager = new UserManager({
+      authority: 'https://192.168.99.100:8443/auth/realms/hmda',
```
hkeeler: Same `UserManager` question as above.  This one does seem a bit different, though.  For instance, there's no _silent renew_-related attributes.  Is that intentional?  What's the difference between these two `UserManager`s?

When we do change this, could we create a JSON file that gets read in instead so we don't have to find/replace in-code?  Also, then it'd be setup in a single place.

src/js/reducers/index.js
```diff
@@ -54,8 +55,24 @@ const defaultEdits = {
   groupByRow: false
 }
 
+/*
```
hkeeler: Should either remove or comment why it's still here.

hkeeler (commented):
All reviewed up.  No show-stoppers.  Mostly questions and thoughts on future improvements.

src/js/UserManager.js
wpears: How is the value passed in, then? Environment variable, some discovery endpoint, etc?

src/js/api.js
wpears: Probably never. That was just put in place as a sane-ish default (which might actually be worse because it will be less obvious what's happening in case of failure)

src/js/UserManager.js
hkeeler: Yes, envvars.

package.json
```diff
@@ -11,9 +11,9 @@
...
+    "js": "browserify -d -t [ babelify --presets [ es2015 react ] --plugins [ babel-plugin-transform-object-rest-spread ] ] -t [ envify --HMDA_API https://192.168.99.100:4443/hmda ] -p [ minifyify --map app.min.js.map --output ./dist/js/app.min.js.map --uglify [ -cm ] ] ./src/js/index.js > ./dist/js/app.min.js && browserify -d -t [ babelify --presets [ es2015 react ] --plugins [ babel-plugin-transform-object-rest-spread ] ] -p [ minifyify --map silent_renew.js.map --output ./dist/js/silent_renew.js.map --uglify [ -cm ] ] ./src/js/silent_renew.js > ./dist/js/silent_renew.js",
     "sass": "node-sass --quiet --output-style=compressed --source-map-embed --source-map=./dist/css/ ./src/sass/app.scss ./dist/css/app.min.css && cp node_modules/uswds/src/fonts/sourcesanspro-* ./dist/fonts",
-    "html": "html-minifier --collapse-whitespace ./src/index.html -o ./dist/index.html",
+    "html": "html-minifier --collapse-whitespace ./src/index.html -o ./dist/index.html, && html-minifier --collapse-whitespace ./src/silent_renew.html -o ./dist/silent_renew.html",
```
awolfe76: Minor but I think you can remove the `,`  before the `&&` from `./dist/index.html, &&`.

src/js/api.js
hkeeler: Yeah, seems like it'd be a bit of a head-scratcher if it gets into that state.  Perhaps log an error instead of setting the default?

awolfe76: Is it from https://github.com/cfpb/hmda-platform-auth/pull/19/files#diff-eb672b1ddd3aa7dae5f2577f38daa271R48?

src/js/UserManager.js
wpears: I'll hold this for now, I think it would make a lot of sense to do when we switch to `hmda-platform`'s docker-compose

src/js/containers/Login.jsx
wpears: Yeah this was actually a holdover from a different way of modeling the state that I overlooked. Removed in subsequent commit.

src/js/UserManager.js
hkeeler: Cool by me.  Added it to https://github.com/cfpb/hmda-platform/issues/661#issuecomment-257983236.

src/js/api.js
```diff
@@ -1,26 +1,37 @@
 import fetch from 'isomorphic-fetch'
 
-function sendFetch(suffix, options = {method: 'GET'}){
-  var url = makeUrl(parseLocation(), suffix);
+let accessToken
 
+function sendFetch(suffix, options = {method: 'GET'}){
+  var location = options.noParse ? {} : parseLocation();
```
awolfe76, line 6: Not sure I'm following whats going on with the `noParse` option.

hkeeler: Same.

wpears: The url for the fetch is generally built out of the url of the current page. However, there are some cases where this isn't true, and in those cases the `location` shouldn't be parsed and used for path parts.

awolfe76: So why is that the case for the `/institutions` calls and nothing else?

wpears: So `<server>/1/2017` can be mined for the institution id and the filing period to use in api requests, but a request for individual institutions from the homepage or institutions page can't use anything in the url (this is why the function signature includes the `id` param for `getInstitution`)

src/js/api.js
```diff
@@ -1,26 +1,37 @@
...
+  var url = makeUrl(location, suffix);
   if(typeof options.body === 'object') options.body = JSON.stringify(options.body)
+  var headers = {}
 
+  if(accessToken) headers.Authorization = 'Bearer ' + accessToken
+  console.log('requesting with access token: ', accessToken)
   var fetchOptions = {
-    method: options.method,
+    method: options.method || 'GET',
```
awolfe76, line 13: Do we need the `|| 'GET'` since its the default?

src/js/api.js
```diff
@@ -1,26 +1,37 @@
...
+  return accessToken || ''
 }
 
 export function makeUrl(obj, suffix){
-  var url = location.protocol + '//' + location.host + '/api'
+  var url = process.env.HMDA_API
+  if(!url) throw new Error('No url provided for API, unable to fetch data. This is most likely a build issue.')
   if(obj.id) url+= '/institutions/' + obj.id;
   if(obj.filing) url+= '/filings/' + obj.filing;
   if(obj.submission) url+= '/submissions/' + obj.submission;
```
awolfe76, line 36: Its not part of this PR but is `obj.submission` ever set?

hkeeler (approved):
All my auth integration related issues have been resolved or moved to other issues.  @awolfe76, once you're happy with it, merge away.

awolfe76:
__We should hold off merging this!__
Will explain more tomorrow at standup.

src/js/api.js
wpears: Aha! I thought the same, but if you pass in your OWN object, the default won't exist! You are setting the _object_ as the default, not the key.

When I first wrote it, I thought the same as you, perhaps this means using a default argument here is worse because of the confusion... that said, who doesn't like default args

src/js/api.js
wpears: Not anymore, and you can see that generally the submission function calls include `/submissions/` in their suffixes. I didn't drop it from the code (yet) because I thought we might want submission id in the route at some point. If you'd prefer, I could nix it and just re-add it if we need it later.

awolfe76: Maybe just comment it out with a note.

awolfe76:
Closes #227 and closes #244

src/js/api.js
awolfe76: Cool. Its working so 💯 . We can refactor later to make it a bit more clear.

awolfe76 (approved):
Got it running locally!

PR #268, opened 2016-11-23
Institutions

awolfe76:
- updates institutions listing page to be a bit more friendly
  - the `status.message` is color coded, green for completed, red for not started and cancelled, and blue for in progress
  - the description text is modified based on the `status.code` too
- adds [moment.js](http://momentjs.com/) to handle time rendering
- adds buttons based on status, with the refile button always present but smaller
- adds column to grid with placeholder text, we can add help text or other messages there
- would like to know if you have thoughts/suggestions on the [`renderTiming`](https://github.com/cfpb/hmda-platform-ui/compare/institutions?expand=1#diff-86bed846ff53cc9ec2c51bcb085621d4R6) and [`renderStatus`](https://github.com/cfpb/hmda-platform-ui/compare/institutions?expand=1#diff-86bed846ff53cc9ec2c51bcb085621d4R26) functions

#### Screenshot
hmda-fis

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,10 +88,11 @@
   "dependencies": {
     "babel-polyfill": "6.9.1",
     "babelify": "7.2.0",
-    "immutable": "3.8.1",
     "envify": "3.4.1",
+    "immutable": "3.8.1",
     "jquery": "3.1.0",
     "jquery.easing": "1.4.1",
+    "moment": "2.17.0",
     "oidc-client": "1.2.0",
     "react": "15.3.0",
     "react-dom": "15.3.0",
--- a/src/js/components/Institutions.jsx
+++ b/src/js/components/Institutions.jsx
@@ -3,25 +3,24 @@ import { Link } from 'react-router'
 import UserHeading from '../components/UserHeading.jsx'
 import moment from 'moment'
 
-const renderLabel = (status) => {
-  let bgColor = 'bg-color-red'
-  if(status.code === 2) {
-    color = ''
+const renderTiming = (status, start, end) => {
+  // default to code 1, not-started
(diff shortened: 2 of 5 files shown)


wpears:
Cool changes.

I think there needs to be more hierarchy/spacing within a bank's filings though (see attached). Also, I think the filing period absolutely needs to be called out in the filing header and perhaps the repetition of the institution name is redundant.

Also, looks like the `start` and `end` fields are 0 when the filing hasn't been started and this isn't being accurately dealt with in `renderTiming`

screen shot 2016-11-28 at 3 11 51 pm

awolfe76:
> I think there needs to be more hierarchy/spacing within a bank's filings though (see attached).

We wouldn't have this case though would we? When someone logs in they would only see the current year. For now, this is only 2017. In the future we would provide some way to go to previous years, maybe a drop-down. I think https://github.com/cfpb/hmda-platform-ui/issues/269 helps here too.

> Also, looks like the start and end fields are 0 when the filing hasn't been started and this isn't being accurately dealt with in renderTiming

Fixed in https://github.com/cfpb/hmda-platform-ui/pull/268/commits/8b00fc953ddb9f40c33e2fc93db3f3ec660467d7

> perhaps the repetition of the institution name is redundant.

Updated in https://github.com/cfpb/hmda-platform-ui/pull/268/commits/81b8276a7c99050241a73f31370403423a9709cd, but also think that we'll improve this even more very soon.

Also added the FI ID (Name - ID), we can figure out what the best ID is later to add to the API.
And changed the position of the buttons, this seems like a better flow.

#### Updated Screenshot

hmda-fi

PR #279, opened 2016-12-06
Replace envvars at runtime

### LINKED ISSUE #270: Replace hardcoded URIs with envvars at container startup
Originally discussed on https://github.com/cfpb/hmda-platform-ui/pull/257#discussion_r86212998, there are currently hardcoded (`http(s)://192.168.99.100:<port>`) URIs in the frontend for:

* HMDA API
* Keycloak
* OIDC-related redirects

These should be set/replaced at container startup via envvars set at container startup.

This will have to be resolved before the frontend will work in a prod-like env.

wpears:
The following URIs are used:
`APP_URL` for the expected uri for the homepage, defaults to `http://192.168.99.100`. Redirect callbacks are constructed from this endpoint.
`HMDA_API` for the expected api endpoint, defaults to `https://192.168.99.100:4443/hmda`
`KEYCLOAK_URL` for the expected keycloak endpoint, defaults to `https://192.168.99.100:8443/auth/realms/hmda`

Closes #270


awolfe76 (approved):
👍

hkeeler:
Looks good.  Will there be a corresponding PR for the changes to `docker-compose.yml`?  I realize there are defaults that fall back to usual docker host IP if the envvars aren't set, but I think it'd be useful to have them Compose file too.

awolfe76:
@wpears - see @hkeeler's comment but go ahead and merge when you let him know.

Issue #269, opened 2016-11-30
Call /institutions/<institution>/filings/<period> for institutions listing


awolfe76:
It seems like calling `/institutions/<institution>/filings/<period>` instead of `/institutions/<institution>` would work better for the institutions listing. We would need to set the `<period>`, initially only 2017, but that endpoint provides more information about a specific year ... filing, submissions, etc. Using it would allow us to render the everything we need, including links to previous submissions within a filing year.

PR #292, opened 2016-12-08
Filing year

wpears:
Depends on #279 

Closes https://github.com/cfpb/hmda-platform-ui/issues/269

Reworks how filings are modeled in the store and fires requests to the `filings` endpoint to populate it along with institutions.


awolfe76:
@wpears - I'm seeing `TypeError: Cannot read property 'institutionId' of undefined(…)` when loading the institutions page. One of 2 may load and others not. It seems to be inconsistent.

(I've gone into the Keycloak admin and added my user to `0,1,2,3`.)

awolfe76:
@wpears - just a general question. I still see we're calling `/institutions` to get the full list, ok there. But then we're calling `/institutions/<institution>`  and `/institutions/<institution>/filings/<period>`. 

Do we still need the call to `/institutions/<institution>`? I was hoping we could remove it because the other 2 endpoints get us everything we need. _I think_.

wpears:
Nope, the filings endpoint doesn't have institution name, which is only returned from the institutions endpoint.

awolfe76:
Right, the `/institutions` has it. So is there anything that `/institutions/<institution>` is getting us?

wpears:
`/institutions/<institution>` allows us to discover the available filing periods for each institution, so we can request only the `/institutions/<institution>/filings/<filing>` that matches the current filing period.

We talked a bit about whether there would always be a default filing for each period, etc, but right now as it stands, just constructing filings endpoints from the info in `/institutions` would result in 404s for institutions that don't have filings for the current filing period.

An alternate approach would be to just intercept 404s and assume they indicate that no filing exists for that institution in the current filing period, but that seems a bit heavy-handed to me.

awolfe76:
I think the 404 is an ok response, especially given that scenario is very unlikely in the future because, like you said, we'll have the default filing period there. That would save us from additional API calls too.

src/js/actions/index.js
```diff
@@ -448,13 +461,50 @@ export function fetchEachInstitution(institutions) {
...
  * Fetch an institution via the api and dispatch an action with the results
  */
 export function fetchInstitution(institution) {
   return dispatch => {
     dispatch(requestInstitution())
     return getInstitution(institution.id)
-      .then(json => dispatch(receiveInstitution(json)))
+      .then(json => {
+        dispatch(receiveInstitution(json))
+          console.log('filings for dispatch', json,json.filings)
```
awolfe76, line 486: Will you still need the logging here?

src/js/api.js
```diff
@@ -23,7 +23,11 @@ function sendFetch(suffix, options = {method: 'GET'}){
 
   console.log('fetching from', url, 'with options', fetchOptions)
   return fetch(url, fetchOptions)
-    .then(response => {console.log('respose from fetch', response);return response.json()})
+    .then(response => {
+      const json = response.json()
+      console.log('respose from fetch', response, json)
```
awolfe76, line 28: Logging again

src/js/components/Institutions.jsx
```diff
@@ -90,30 +90,32 @@ const renderRefile = (code, institutionId, period) => {
...
+  for(let i=0; i<institutions.length; i++){
+    if(institutions[i].id === filing.institutionId) return institutions[i]
+  }
+  return null
+}
+
 export default class Institution extends Component {
   render() {
-    var self = this
+    console.log('inst filings',this.props.institutions, this.props.filings)
```
awolfe76, line 102: Logging

awolfe76 (approved):
Just the few comments about logging.

PR #294, opened 2016-12-15
Navigation and refiling

wpears:
Introduces a (very minimally designed) nav which is aware of the app's state and works with refiling.

Known issues:
 - Nav shouldn't be straight links, or rather, only links if those parts have been completed (along with some sort of affordance to show what has been completed or not)
- the links to next items "review edits" etc, should really be a more fully fledged solution, possibly baked in with the refilewarning and/or validation progress. If not baked, they should at least seem cohesive and point the user in the same direction


src/js/components/Institutions.jsx
```diff
@@ -87,9 +87,10 @@ const renderButton = (code, institutionId, period) => {
   return <Link className="usa-button" to={`/${institutionId}/${period}`}>{buttonText}</Link>
 }
 
-const renderRefile = (code, institutionId, period) => {
-  if (code === 1) return null
-  return <Link className="usa-button usa-button-secondary usa-text-small" to={`/${institutionId}/${period}`}>Refile</Link>
+const renderRefile = (makeNewSubmission, code, institutionId, period) => {
+  if(code === 1) return null
+  return <a className="usa-button usa-button-secondary usa-text-small" onClick={()=>{
```
awolfe76, line 92: Is there a reason to not use a `<Link>` here?

wpears: Cause all we care about is the action that the click triggers, not the special behavior of the Link component.

awolfe76 (approved):
👍  Also created #296 for later.

PR #295, opened 2016-12-15
Signature submit

awolfe76:
Depends on #293 (and probably #294)

- adds submit button to signature
- adds action/reducer/constant for the checkbox
  - adds a `checked` property to the state for the checkbox
- checkbox no longer uses dispatch for API call, the button does


awolfe76:
@wpears - this one should be ready for review now too.

wpears (approved):

PR #307, opened 2016-12-30
Test architecture

awolfe76:
__This will probably cause conflicts with other PRs. Should probably think about when to merge it.__

This is to make it easier to determine what needs tests and what already has them by structuring the `__tests__` the same as the `src`.

```
Wrapper.js
actions/
components/
containers/
reducers/
```


wpears:
Makes sense to me; i'd say we should get the other PRs merged in first though, then work through any conflicts in the PR to polish it off.

awolfe76:
This one should be ready @wpears.

PR #309, opened 2017-01-05
Implement single quality verify checkbox without undo

wpears:
Needs some fixup on tests and probably shouldn't merge until the backend finalizes https://github.com/cfpb/hmda-platform/issues/747 so the api can be real.

I'm fairly uncomfortable with the not-being-able to uncheck thing; seems like the messaging for refiling quality edits should be earlier (say, right before quality edits or in the header). A locked-in checkbox kinda breaks user expectations (speaking from my own perspective as I tried it out). This might make the  backend call slightly more complex (actually needs a boolean), but I think it's worth it.


awolfe76:
@wpears  - I agree.

> seems like the messaging for refiling quality edits should be earlier (say, right before quality edits or in the header)

Yeah, let's add a message in the header like we do for others. Probably a change to `components/RefileWarning` to make it show more than one message.

Also, I think we can show a message when the checkbox is unchecked. A similar message (if not exactly the same) to the one in the header, only this time just above or below the checkbox.

> This might make the backend call slightly more complex (actually needs a boolean), but I think it's worth it.

It should be a boolean on the back-end for sure. The 2nd part of https://github.com/cfpb/hmda-platform/issues/747#issuecomment-268373672 talks about the `GET` which has true or false. (Maybe that issue needs some more clarification?)

All of this ends up making quality edits act more like syntactical and validity. It will prevent a signature from being added and require a re-upload if something is wrong ... in the case of quality we're just providing a way to verify.

wpears:
@awolfe76 merge conflicts resolved and quite a few nav-related goodies added in (enhancements to both the editsnav and the button to advance to the next page).

Also includes logic to both to summary page when applicable (checking whether everything has been verified/justified).

The unmocked call to the verify quality endpoint is still blocked by https://github.com/cfpb/hmda-platform/pull/800, but I assume that will be rectified posthaste

awolfe76:
I think we wanted filers to be able to view all the edit, but maybe not interact with them, based on the status code. That way they could at least see everything.

wpears:
Yeah, I was a big proponent of that model, but I think in practice it makes the app feel more sprawling / less sensical. I think it also makes the "hey you actually have to refile now" call to action less clear and would require more info on the quality/macro pages so that people know what's going on when they click through but can't interact.

My original idea was that we should be minimizing the total number of uploads, but now I think we should be minimizing the amount of possible confusion at every stage in the process. Also, I think in practice this would result in <1 extra upload per filer while making the process / required interaction much clearer.

wpears:
@awolfe76 the quality endpoint is in master on the platform and the call has been implemented here. 🚀

awolfe76 (approved):
Awesome! 👍

PR #310, opened 2017-01-06
CSV download and updates to match the edit API

### LINKED ISSUE #280: Remove separate rendering of transmittal sheet
Depends on https://github.com/cfpb/hmda-platform/issues/721

If we update the response in https://github.com/cfpb/hmda-platform/issues/721 we can remove [this function](https://github.com/cfpb/hmda-platform-ui/blob/master/src/js/containers/EditsTable.jsx#L24) and [the call to it](https://github.com/cfpb/hmda-platform-ui/blob/master/src/js/containers/EditsTable.jsx#L57).

### LINKED ISSUE #308: Implement actual previous submission download
Links now exist in the UI, but they need to function as expected

awolfe76:
~~__WIP - failing tests but wanted to have you look at a few things first ... see below__~~

Closes #308 and closes #280 

- allows for csv download on institutions listing page
  - would really like you to check out the changes to `api.js`, it feels a bit off
  - and there's a lot of arguments being passed around through the `institutions` container and component to get there

Also

- due to changes in https://github.com/cfpb/hmda-platform/pull/760 updates how tables, rows, and cells are rendered for edits
  - and cleans it up a bit to use the `type` instead of looking at the `object` to determine what to render

### DIFF
--- a/src/js/actions/index.js
+++ b/src/js/actions/index.js
@@ -535,8 +535,6 @@ export function fetchEditsByType() {
     dispatch(requestEditsByType())
     return getEditsByType(latestSubmissionId)
       .then(json => {
-        console.log('fetchEditsByType')
-        console.log(json)
         dispatch(receiveEditsByType(json))
       })
       .catch(err => console.error(err))
--- a/src/js/actions/index.js
+++ b/src/js/actions/index.js
@@ -112,7 +112,6 @@ export function requestEditsByRow() {
 }
 
 export function receiveEditsByType(data) {
-  console.log('edits by type', data)
   return {
     type: types.RECEIVE_EDITS_BY_TYPE,
     edits: data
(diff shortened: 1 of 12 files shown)


awolfe76:
@wpears - thanks for the help on the test! It should pass this time and this is good for review.

src/js/api.js
```diff
@@ -2,8 +2,24 @@ import fetch from 'isomorphic-fetch'
 
 let accessToken
 
+function createQueryString(suffix, params, length) {
```
wpears: I'd prefer this as 

```
function createQueryString(suffix, params) {
  const length = Object.keys(params).length
  let newSuffix = `${suffix}?`
  let count = 0
  for(let key of Object.keys(params)) {
    newSuffix = `${newSuffix}${key}=${params[key]}`
    count++
    if(count < length) newSuffix = `${newSuffix}&`
  }
  return newSuffix
}
```

Since the length will always be calculated the same way

src/js/api.js
```diff
@@ -2,8 +2,24 @@ import fetch from 'isomorphic-fetch'
...
+  return newSuffix
+}
+
 function sendFetch(suffix, options = {method: 'GET'}){
   var locationObj = options.noParse ? {} : parseLocation(location);
+  // check if params exist and update suffix to add them
+  if(options.params && Object.keys(options.params).length !== 0) {
+    suffix = createQueryString(suffix, options.params, Object.keys(options.params).length)
+  }
+  console.log(suffix)
```
wpears: Extraneous log

src/js/actions/index.js
```diff
@@ -491,7 +534,11 @@ export function fetchEditsByType() {
   return dispatch => {
     dispatch(requestEditsByType())
     return getEditsByType(latestSubmissionId)
-      .then(json => dispatch(receiveEditsByType(json)))
+      .then(json => {
+        console.log('fetchEditsByType')
+        console.log(json)
```
wpears: Logs

src/js/api.js
```diff
@@ -91,8 +117,11 @@ export function getLatestSubmission(){
   return sendFetch('/submissions/latest')
 }
 
-export function getEditsByType(submission){
-  return sendFetch(`/submissions/${submission}/edits`)
+export function getEditsByType(submission, institutionId, period, params){
+  if(params) {
+    return sendFetch(`/institutions/${institutionId}/filings/${period}/submissions/${submission}/edits`, {noParse:1, params:params})
+  }
+  return sendFetch(`/submissions/${submission}/edits`, {params:params})
```
wpears: The last line of this fn can just be
```
return sendFetch(`/submissions/${submission}/edits`)
```

src/js/api.js
```diff
@@ -24,6 +46,10 @@ function sendFetch(suffix, options = {method: 'GET'}){
   console.log('fetching from', url, 'with options', fetchOptions)
   return fetch(url, fetchOptions)
     .then(response => {
+      if(options.params && options.params.format === 'csv') {
+        //console.log(response.text())
+        return response.text()
+      }
       const json = response.json()
       console.log('respose from fetch', response, json)
```
wpears: logs can be removed

src/js/actions/index.js
```diff
@@ -293,6 +292,50 @@ export function fetchSummary() {
   }
 }
 
+// used to trigger csv download properly
+function detectIE() {
+  const ua = window.navigator.userAgent;
+
+  const msie = ua.indexOf('MSIE ');
+  if (msie > 0) {
```
wpears: I'm a little hesitant of parsing the user agent, but I'll allow it if you turn the comparison operator to `>=` since this would fail if the matching string in a new ie version happened to start at position 0

PR #312, opened 2017-01-06
Fix js sourcemap

awolfe76:
When rebuilding the js while in a running container browsers aren't able to load the js sourcemap.

It _seems_ like removing the `app.min.js.map` completely before the rebuild lets the browsers find it again.

Give it a shot.


wpears:
If this was the only command you ran (rather than manually removing the source map file), I doubt that that was the problem, since there is a typo in `rimraf .dist/js/app.min.js.map`. Could be that changing files/generating a new source map simply requires you to open a new tab in Chrome and that you did so without noticing (that seems to typically clear up any source map woes I've come across with the current setup)

awolfe76:
Sorry, that was probably a typo made when creating this branch (had it in another branch to begin with). I just tried it again and it still seems to work. I've done several rebuilds and refreshing each time keeps the sourcemap. Fixed with 0949eef.

I tried the new tab too but that didn't seem to work. Hmmm.

wpears:
I feel like this might be browser version related or in someway external to the project. That said, I'll merge this as it doesn't seem to affect me but fixes your setup. I'd say keep an eye out for this issue cropping up later though as I'm not sold that this is the root cause.

PR #319, opened 2017-01-12
Implements sort by row

### LINKED ISSUE #272: Allow sorting by row
Depends on https://github.com/cfpb/hmda-platform/issues/718

- [ ] add button to UI to allow "sort by row"
  - [ ] this button should change between "sort by row" and "sort by edit"
- [ ] update API call depending on what is selected (defaults to "by edit")
  - if "by row" is chosen add query string to the API call `sortBy=row`

A lot of the code for this is in place, once the back-end is ready we'll need to test and make any updates needed.

wpears:
- rewrites api.js a bit to better handle the `getEdits` endpoint
 - reorganizes EditsByType and EditsByRow to EditsTableWrapper for better code reuse
 - implements SortPicker component to switch between grouping by row or edit type

Closes #272


src/js/components/SortPicker.jsx
```diff
@@ -0,0 +1,33 @@
+import React, { PropTypes } from 'react'
+
+const getText = (groupByRow) => {
+  if(groupByRow) return 'Group by Edit'
+  return 'Group by Row'
+}
+
+const getClass = (groupByRow) => {
```
awolfe76: Any reason to change the color of the button in this case? I don't think the button should change other than the text. If we want to make this more apparent we could have both options present all the time, highlighting the current selection with the [`active` state](https://standards.usa.gov/buttons/).

awolfe76, line 3: Any reason to change the color of the button in this case? I don't think the button should change other than the text. If we want to make this more apparent we could have both options present all the time, highlighting the current selection with the [`active` state](https://standards.usa.gov/buttons/).

awolfe76 (commented):
Other than the one comment, 👍 .

src/js/components/SortPicker.jsx
wpears: Yeah, I think I like that more. I'll drop that in

awolfe76:
@wpears - looks nice! But travis is failing. Once that's fixed you can merge.

PR #323, opened 2017-01-18
Edits csv download

awolfe76:
- adds in CSV download for each edit type on the `/edits` "page"
  - uses `javascript:void(0)` as the `href` (for the CSV download on the `/institutions` too) ... is there a better way?
- quick fix for edit count always saying `0`

### DIFF
--- a/__tests__/components/EditsHeaderDescription.js
+++ b/__tests__/components/EditsHeaderDescription.js
@@ -22,7 +22,7 @@ describe('EditsHeaderDescription', function(){
   })
 
   it('correctly sets the syntactical desc', function(){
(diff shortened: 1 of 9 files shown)


wpears:
Yeah, I'd say href='#' and `event.preventDefault()`ing is probably a better solution. I think the entire `window.open` scheme can actually be improved upon as well;; working on something along these lines

wpears:
I dropped in the `preventDefault` change and also added the `file-saver` module which will allow us cross-browser downloading logic without having to roll it ourselves.

awolfe76:
Nice! Those changes look good to me.

PR #332, opened 2017-01-23
Upload form styles

awolfe76:
__Depends on #331__

- clean up layout for form
- disables upload button after submission
- make review edits a button

#### Before upload
upload

#### After upload
uploaded


awolfe76:
@wpears - this is ready now.

#### New screenshot after upload
uploaded

Issue #290, opened 2016-12-07
Don't allow verification of edits or signature based on state


awolfe76:
We'll be showing filers the all edits but syntax and validity require a re-filing.

- [x] do not allow verification of quality and macro if syntax and validity edits exist. (#351)
- [x] do not allow signature if status < 12 (this will be 11 after https://github.com/cfpb/hmda-platform/issues/729)

PR #351, opened 2017-01-26
Disable macro

awolfe76:
Closes part of #290 

- enable macro justifications when not signed and there are no syntactical and validity edits

To visually test you can use the file in https://github.com/cfpb/hmda-platform/pull/793.


awolfe76:
@wpears - i think this one is ready

PR #355, opened 2017-02-01
Confirm resubmission process

### LINKED ISSUE #161: Resubmission alert
Alert the filer that the loan/application register for the reporting year has been filed, and prompt the filer to confirm that he wishes to refile the loan/application register.

awolfe76:
Closes #161 

- adds a confirmation message if they refile now button is clicked on /upload

#### Screenshot
are-you-sure


wpears:
I think this kinda circumvents the redux state and standard action/reduction flow.

I'm going to add a commit in using this model to see what you think

awolfe76:
It definitely does a bit but I thought since showing that confirmation message doesn't rely on any part of the existing state and it doesn't need to be "saved" (its really only needed for this component) this would be ok.

Happy to see a different approach though.

PR #357, opened 2017-02-01
set the table column widths

### LINKED ISSUE #356: Set table column widths
The column widths are uneven when viewing the edits.

Example:
table-widths

awolfe76:
Closes #356 

- set the widths based on the number of fields


wpears (approved):
:+1:

PR #360, opened 2017-02-02
WIP - Home page

### LINKED ISSUE #278: Create "static" home page
Our application should have a "static" home page when someone first comes to the application.

Possible content:
- FAQs (#277)
- header (#230)
  - login
  - ffiec logo
  - navigation (TBD)
- footer (#231)
- other messages we need (alert during filing period for example)
- links to docs (API, reg, help desk, other guides, etc)

awolfe76:
_I marked this as __WIP__ but we can merge this or just leave it for a bit until we get the real content. I'm good either way._

Closes #278 

- sets up the basic framework for the home page
- placeholders are just examples of things we can use but can be removed
- has an example of different content if we have a `user` 
  - could also do the same thing based on key dates

---

#### Screenshot (not logged in)
home-page-2

#### Screenshot (logged in)
home-page

### DIFF
--- a/src/js/components/Home.jsx
+++ b/src/js/components/Home.jsx
@@ -1,5 +1,23 @@
 import React from 'react'
 import Header from '../components/Header.jsx'
+import { Link } from 'react-router'
+import { signinRedirect } from '../redirect.js'
+
+const getLoginMessage = (userName) => {
+  if(userName) return null
+
+  return (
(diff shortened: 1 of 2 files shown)


awolfe76:
@cfpb/hmda-team-admin - if anyone has ideas about content for the home page just update this PR.

PR #361, opened 2017-02-02
remove unused return

awolfe76:
@wpears - I just noticed this after the merge. There was a 2nd `return` that I had in there while testing, this removes it.


wpears (approved):
Hah I thought that was a weird artifact of the diff ;)

PR #363, opened 2017-02-07
Modal confirm

wpears:
Needs tests, but give it a whirl.

Refiling from anywhere in the app will trigger the modal, which will clear upon selecting Yes/No (with the refile actually being triggered by selecting 'Yes').

### DIFF
--- a/src/js/components/ConfirmationModal.jsx
+++ b/src/js/components/ConfirmationModal.jsx
@@ -1,0 +1,25 @@
+import React from 'react'
+
+const ModalConfirm = (props) => {
+  return (
+    <div className={'confirmation-blurred-blocker'+ (props.showing ? ' showing-blurred-blocker' : '')}>
+      <div className="confirmation-modal">
+      <div>
+        <h4>Are you sure?</h4>
+        <button onClick={(e)=>{
+          e.preventDefault()
+          props.hideConfirmModal()
+          props.triggerRefile(props.id, props.filing)
+        }}>Yes</button>
+        <button className="usa-button usa-button-secondary"
+          onClick={(e)=>{
+            e.preventDefault()
+            props.hideConfirmModal()
+          }}>No</button>
+        </div>
+      </div>
+    </div>
+  )
+}
+
+export default ModalConfirm
--- a/src/js/components/ConfirmationModal.jsx
+++ b/src/js/components/ConfirmationModal.jsx
@@ -1,11 +1,12 @@
 import React from 'react'
+import RefileText from './RefileText.jsx'
 
 const ModalConfirm = (props) => {
   return (
     <div className={'confirmation-blurred-blocker'+ (props.showing ? ' showing-blurred-blocker' : '')}>
       <div className="confirmation-modal">
       <div>
-        <h4>Are you sure?</h4>
+        <RefileText code={props.code}/>
         <button onClick={(e)=>{
           e.preventDefault()
           props.hideConfirmModal()
--- a/src/js/containers/ConfirmationModal.jsx
+++ b/src/js/containers/ConfirmationModal.jsx
@@ -10,7 +10,6 @@ class ConfirmationModalContainer extends Component {
   }
 
   render() {
-    console.log(this.props)
     return <ConfirmationModal {...this.props} />
   }
 }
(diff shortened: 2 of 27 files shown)


src/js/containers/ConfirmationModal.jsx
```diff
@@ -0,0 +1,38 @@
...
+import { hideConfirm, createNewSubmission} from '../actions'
+import ConfirmationModal from '../components/ConfirmationModal.jsx'
+
+class ConfirmationModalContainer extends Component {
+  constructor(props) {
+      super(props)
+  }
+
+  render() {
+    console.log(this.props)
```
awolfe76: Logging.

src/js/components/ConfirmationModal.jsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react'
+
+const ModalConfirm = (props) => {
+  return (
+    <div className={'confirmation-blurred-blocker'+ (props.showing ? ' showing-blurred-blocker' : '')}>
+      <div className="confirmation-modal">
+      <div>
+        <h4>Are you sure?</h4>
```
awolfe76: I think it would be nice to give them some indication of the current progress, especially if has been signed ... eg "Your current submission has been signed. Are you sure you want to refile?"

wpears: Agreed, I'll key into the same logic you're using for the institutions page warning

src/sass/components/ConfirmationModal.scss
```diff
@@ -0,0 +1,35 @@
...
+}
+
+.confirmation-modal {
+  width: 30%;
+  min-width: 320px;
+  height: 200px;
+  margin: auto;
+  position: fixed;
+  z-index:3;
+  background:#e6e7e9;
```
awolfe76: You can use something like `$color-gray-light` here instead (or `$color-gray-warm-light`).

src/sass/components/ConfirmationModal.scss
```diff
@@ -0,0 +1,35 @@
...
+  z-index: 2;
+}
+
+.confirmation-modal {
+  width: 30%;
+  min-width: 320px;
+  height: 200px;
+  margin: auto;
+  position: fixed;
+  z-index:3;
```
awolfe76: If you bump `z-index` to at least `9001` the nav stuff will be under the overlay too. The `.usa-nav` in USWDS has a `z-index` of `9000` which is why it's not.

src/sass/components/ConfirmationModal.scss
```diff
@@ -0,0 +1,35 @@
...
+  background:#e6e7e9;
+  border-radius: 4px;
+  box-shadow: 0 0 4px -1px #525652, 0 1px 2px -1px #525652;
```
awolfe76: For the color here you could try `$color-gray` or `$color-gray-warm-dark`.

src/sass/components/ConfirmationModal.scss
```diff
@@ -0,0 +1,35 @@
...
+}
+
+.confirmation-modal > div {
+  width: 145px;
+  margin: 50px auto 0;
+}
+
+.confirmation-blurred-blocker {
+  background: rgba(255,255,255,0);
```
awolfe76: This is picky, but the base color for black is `rgb(33,33,33)` in USWDS if you want to use that instead.

wpears:
@awolfe76 should be good for another once over. On the upload page I moved the refile button to the top left.. kinda trying it out as a permanent position for that button across all the submission pages (maybe not a part of this PR, just wanted to hear your thoughts).

Custom refile message based on submission status code is now rendered in the modal (either in progress or complete).

awolfe76:
The only things I see are, after signing the submission the alignment is off:
refile-after-sign

And the refile button is on the upload "page" when you first get there, both at the start (first submission) and after clicking the refile button:
refile-button-showing

awolfe76:
I'm ok with having the re-file button always there, but in that case maybe we remove it from the re-file warning? If you look at a file that fails parsing you'll see the button twice. So if its in the refileWarning component it would show up twice.

Example:
refile-twice

wpears:
Yeah, I didn't add it in yet/make the necessary changes cause I wanted your input first ;D

awolfe76 (approved):
👍

PR #378, opened 2017-02-15
CSS refactor

awolfe76:
More to come but I think this is a good start. This is mostly to clean out the `app.scss` file. I'm going to go to each component next to see if we can improve a few things.

- use `<hr>` and `<header>` where appropriate
- clean up `app.scss` by removing unused styles and moving styles to the components
- create a few `<Component>.scss` files
- some other refactoring on components


wpears (approved):
🌠

Issue #380, opened 2017-02-15
Refiling doesn't clear previous edits


awolfe76:
If you upload a file with edits, then refile, the edits from the previous submission are still present. A refresh will clear them.

awolfe76:
Closed by #383

Issue #379, opened 2017-02-15
"View filing" button on institution page bugged


wpears:
repro: Start filing, navigate to the institution page, click "view filing" on the in-progress filing

awolfe76:
Had this in my notes too 😄

awolfe76:
Closed by #383

Issue #367, opened 2017-02-10
"Begin filing" button on home page


wpears:
Clear, large call to action maybe it's "Login and begin filing" but it redirects to the institutions page which will automatically kick to the login page if the user isn't logged in

awolfe76:
Closed by #383

PR #383, opened 2017-02-16
Postrelease bugfix

### LINKED ISSUE #381: Edits nav links showing before they should
If you upload a file sometimes the nav renders with links instead of "grayed out".

For example, if you upload `failsMacro.txt` the "Summary" text will render as a link but it shouldn't. Same happens sometimes when a file fails syntax and validity edits, the quality and macro links will render.

Maybe related to #380.

wpears:
Closes #381, #380, #379, #367


package.json
```diff
@@ -10,16 +10,18 @@
...
     "postinstall": "npm run clean",
-    "watch:js": "nodemon --exitcrash -e js,jsx -w src/js -x 'npm run js'",
+    "watch:js": "nodemon --exitcrash -e js,jsx -w src/js -i src/js/silent_renew.js -x 'npm run js'",
+    "watch:renew": "nodemon --exitcrash -w src/js/silent_renew.js -x 'npm run renew'",
     "watch:tests": "nodemon --exitcrash -e js -w __tests__ -x 'jest --coverage -o'",
     "watch:sass": "nodemon --exitcrash -e scss -w src/sass -x 'npm run sass'",
     "watch:html": "nodemon --exitcrash -e html -w src -x 'npm run html'",
-    "watch": "npm run watch:js & npm run watch:tests & npm run watch:sass & npm run watch:html",
+    "watch": "npm run watch:js & npm run watch:renew & npm run watch:tests & npm run watch:sass & npm run watch:html",
     "build": "npm run js && npm run sass && npm run html && npm run img"
```
awolfe76: We'll need to upload the `build` with the `renew` too.

awolfe76:
@wpears 
I still see a something off in the edits nav.

Upload `validation_sample.txt`, everything is ok; only the "Syntactical & Validity" renders as a link.
Do the refile and choose the `failsMacro.txt` file. Only the "Syntactical & Validity" renders as a link still. Clicking the link will make "Quality" and "Macro" render as links ... which is correct but it doesn't happen until you are on the `/syntacticalvalidity` page.

The edits are clearing, though. 👍

wpears:
Yeah the reason this is acting up is because the front-end only requests edits when it actually moves to an edits page. Before all the links were rendering because it seemed like there were no edits when the check went off (whereas actually they just hadn't been requested yet).

Also, I think it may actually be the preferred UX to ensure users look at every page in the expected order, so that's why i went with the change of having synval link from the upload page even if there are no edits. It's kinda like passing through a checkpoint.

awolfe76:
Makes sense. Maybe we can come up with something cleaner if we update the states on the back-end. Merging! (after travis finishes)

PR #384, opened 2017-02-16
Submission status

### LINKED ISSUE #303: Update links to previous submissions
Depends on #302 

Update the submissions listings from #298 to:

- [ ] show a link to a csv download for that upload if it is not "completed"(filing status of 3)
- [ ] ~show a link to view the submission in the UI if the status is "completed"~ (see #302)
- [ ] show the previous submission if "completed" without a download

### LINKED ISSUE #382: Overall filing status on institutions listing page
If you complete a filing the institution status is correct, it says completed. But, if you refile (a new submission is created) that status still shows completed.

This is technically correct according to the back-end, the filing status remains completed because the institution has a completed submission.

The UI is misleading in this case though. It feels like it should be in progress.

Maybe the status on the institutions listing page should be based on the latest submission?

awolfe76:
Closes #382 
Closes #303 

- Uses submission on instead of filing to display status on institutions listing
  - prevents seeing odd cases when filing, eg if a filing has been signed by a filer still clicks refile
- highlights the current status vs previous submissions a little better
- show the previous submission statuses
- some general cleanup

submissions


src/js/components/Institutions.jsx
```diff
@@ -109,26 +155,47 @@ export default class Institution extends Component {
...
         userName={this.props.user.profile.name} />
       <div id="main-content" className="usa-grid">
-        <UserHeading period="2017" userName={this.props.user.profile.name} />
+        <UserHeading
+          period="2017"
+          userName={this.props.user.profile.name} />
         <div className="usa-width-two-thirds">
           {this.props.filings.map((filingObj, i) => {
             const filing = filingObj.filing
+            const latestSubmissionStatus = filingObj.submissions[0].status || null
```
wpears: This makes me slightly uncomfortable since I can't find the code in the backend that gives a hard guarantee that the latest submission (highest sequence number) will always be at index 0. I'd like not to burn extra cycles here, but correctness is paramount.

@schbetsy could you weigh on on sorting of submissions?

awolfe76: Yeah, I did have the same thought ... it has been consistently that way but good to ask.

wpears: Following up on this, I filed https://github.com/cfpb/hmda-platform/issues/819

@schbetsy said that under most circumstances, the order will be what we'd expect here, so I'll sign off on this (after sneaking in a guard)

schbetsy: Sorry, GitHub didn't show me this comment until today ( :rage: ). 

But yeah, as @wpears said, the back end just returns these in the order they're persisted--which *should* be the right order--but currently doesn't guarantee an order.

PR #390, opened 2017-02-21
Refactor

### LINKED ISSUE #373: Refactor CSS
- possibly only include what is needed from USWDS rather than `all.scss`
- take advantage of `uswds-customer-vars.scss` where we can
- get back to original idea of using `components/<Component>.scss`
  - could clean out a lot of style from `app.scss`

awolfe76:
Closes #373 

- include only the required USWDS sass files
- refactor NavButton; wording, CSS
- refactor institutions css
- refactor refile button to use a prop for float
- add CSS for sortPicker
- update message for upload progress


src/js/containers/Submission.jsx
```diff
@@ -91,7 +91,11 @@ class SubmissionContainer extends Component {
           period={params.filing}
           userName={user.profile.name}
           institution={params.institution} />
-        {code > 2 ? <div className="FloatingRefile"><RefileButton id={params.institution} filing={params.filing} code={code}/></div> : null}
+        {code > 2 ? <RefileButton
+          id={params.institution}
+          filing={params.filing}
+          code={code}
+          class='float' /> : null}
```
wpears, line 98: This should be `className`, no?

src/js/components/NavButton.jsx
```diff
@@ -11,36 +11,38 @@ const NavButton = (props) => {
...
-      if(code < 8 || syntacticalValidityEditsExist || !qualityVerified || !macroVerified) className+=disabled
+      if(code < 8 || syntacticalValidityEditsExist || !qualityVerified || !macroVerified) className = 'usa-button-disabled'
       break
     default:
       return null
   }
 
-  if(!displayName) displayName = suffix
+  let displayName = (suffix === 'syntacticalvalidity') ? '' : suffix
+  displayName = (suffix !== 'summary') ? `${displayName} Edits` : displayName
```
wpears, line 39: There's an extraneous space here when displayName gets assigned to empty string, but it'll collapse by default.

wpears (dismissed):
Specifically the className thing

src/js/components/NavButton.jsx
awolfe76: In this case it's just a `prop` passed to `RefileButton` and used [here](https://github.com/cfpb/hmda-platform-ui/pull/390/files#diff-06b62e091b18afbc21500ce204115596R4).

src/js/containers/Submission.jsx
awolfe76: In this case it's just a `prop` passed to `RefileButton` and used [here](https://github.com/cfpb/hmda-platform-ui/pull/390/files#diff-06b62e091b18afbc21500ce204115596R4).

PR #391, opened 2017-02-22
Loading indicator

wpears:
Adds a loading indicator to the institutions and submissions pages.

Also includes various state bugfixes and required downstream component updates.


src/js/actions/index.js
```diff
@@ -485,22 +499,14 @@ export function fetchNewSubmission(id, period) {
...
+    return getLatestSubmission()
+      .then(json => {
+        if(json.httpStatus === 404){
+          const splitPath = json.path.split('/')
+          return dispatch(fetchNewSubmission(splitPath[2], splitPath[4]))
         }else{
-          return createSubmission(json.filing.institutionId, json.filing.period).then(submission => {
-              dispatch(receiveSubmission(submission))
-            })
+          return setTimeout(()=>dispatch(receiveSubmission(json)),1000)
```
awolfe76: Why the wait to receive the submission?

src/js/components/Institutions.jsx
```diff
@@ -177,7 +183,13 @@ export default class Institution extends Component {
           period="2017"
           userName={this.props.user.profile.name} />
         <div className="usa-width-two-thirds">
-          {this.props.filings.map((filingObj, i) => {
+          {this.props.isFetching || !this.props.filings ?
+            <div className="usa-grid-full">
+              <p style={{visibility: 'hidden'}}>Loading data</p>
```
awolfe76: Can we just add this to `LoadingIcon`?

src/js/containers/Submission.jsx
```diff
@@ -16,11 +16,46 @@ import Signature from './Signature.jsx'
...
+      if(code > 7){
+        toRender.push(<IRSReport/>)
+        toRender.push(<Summary/>)
+        toRender.push(<Signature/>)
+      }
+    }
+  }
+
+  if(toRender.length === 0){
+    toRender.push(<p>Something is wrong, please <Link to='/institutions'>Go Back</Link></p>)
```
awolfe76: We should say "please go back to your institutions" (or something like that). "Back" might not be that page in all cases, so we should be clear about where we're sending them.

src/js/containers/Submission.jsx
```diff
@@ -35,51 +70,22 @@ class SubmissionContainer extends Component {
...
-    if(this.props.status.code === null){
+    if(!this.props.isFetching &&
+      (!this.props.status || this.props.status.code === 0)){
       this.props.dispatch(fetchSubmission())
-      return null
     }
 
     const status = this.props.status
-    const code = status.code
+    const code = status && status.code
```
awolfe76: These lines, ~80-84, might be cleaner as `const { status, params, user } = this.props`.

src/js/containers/Submission.jsx
```diff
@@ -16,11 +16,46 @@ import Signature from './Signature.jsx'
...
+  const toRender = []
+  if(code === -1) {
+    toRender.push(<p>{message}</p>)
+  }else{
+    if(page === 'upload'){
+      toRender.push(<UploadForm code={code}/>)
+      if(code === 5) {
+        toRender.push(<RefileWarning/>)
+        toRender.push(<ParseErrors/>)
+      }
```
awolfe76, line 35: For the `/upload` page, I think we should add `RefileWarning` if the code is 8 too. Seems like it would be good to let them know on all pages.

src/sass/components/LoadingIcon.scss
```diff
@@ -0,0 +1,34 @@
...
+  transform: rotateX(180deg);
+}
+
+.LoadingIcon {
+  background-image:url('/img/LoadingIcon.png');
+  width: 3em;
+  animation: loading-spin 2s infinite linear;
+}
+
+@-webkit-keyframes loading-spin {
```
awolfe76, line 14: Created #392 to handle this better. I'm sure there are other places too.

awolfe76 (commented):
A few comments, questions but this is great.

src/js/components/Institutions.jsx
wpears: Possibly, not the `<p style={{visibility: 'hidden'}}>Loading data</p>` part alone, cause otherwise the width would collapse. Probably should just adjust the styles for the` LoadingIconWrapper`.

src/js/actions/index.js
wpears: Only for demo purposes. I'll pull this out before merge.

The wait was actually quite helpful in discovering and fixing some latent race conditions

awolfe76 (approved):

PR #405, opened 2017-03-06
Refile

### LINKED ISSUE #333: Trigger resubmission from upload component after first submission
Depends on #161 

In order to keep the upload component as part of the submission flow and breadcrumbs, we'll need to allow uploading a new file to trigger another submission (after warning with a resubmission alert).

This is a little nicer than requiring users to click a specific "refile" link from a different part of the app, since they've already learned where/how they upload a file and shouldn't be required to manually trigger the resubmission if they just manually navigate back to the upload screen and attempt to reupload.

awolfe76:
Closes #333 

- allow upload to happen from upload form (dropzone)
  - adds a new action to track the new file being uploaded
  - still triggers confirmation modal
- remove the "refile" button from edit pages
- adds a few quick test updates


wpears (approved):
Looks good, merging with https://github.com/cfpb/hmda-platform-ui/issues/406 created

PR #411, opened 2017-03-07
updates to help the compose build time

awolfe76:
Steps I took to test the build times:
- removed all containers and images
- ran `time docker-compose up -d --build`

Results:
```
real	11m7.019s
user	0m15.813s
sys	0m5.301s
```

Then:
- made updates to the UI dockerfile
- removed all containers and images again
- ran `time docker-compose up -d --build` again

Results:
```
real	7m50.625s
user	0m15.689s
sys	0m5.264s
```

Looks like it trimmed a little over 3 minutes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,22 +1,13 @@
 FROM nginx:1.10
 MAINTAINER Wyatt Pearsall<Wyatt.Pearsall@cfpb.gov>
 
-RUN apt-get update && \
-    apt-get install -y curl g++ git make && \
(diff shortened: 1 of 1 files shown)


wpears:
The benefit would definitely be felt here, but this essentially makes the dockerfile dev only. Maybe make these changes but keep the original in a file named Dockerfile_prod (which could also be the home of various production settings).

awolfe76:
Our production Dockerfile will come from another source (different repo), so I think we're ok to do this.

wpears (approved):

PR #412, opened 2017-03-09
Pagination for Parse Errors

### LINKED ISSUE #274: Add pagination to a few components
Depends on https://github.com/cfpb/hmda-platform/issues/719

Once the endpoints are returning what's needed we'll have to add pagination to:
- [ ] [`editsTable`](https://github.com/cfpb/hmda-platform-ui/blob/master/src/js/containers/EditsTable.jsx) _(on hold)_
- [ ] parserErrors (depends on #273)
- [ ] [`IRSReport`](https://github.com/cfpb/hmda-platform-ui/blob/master/src/js/components/IRSReport.jsx)_(on hold)_

We'll have to update the api calls to allow for pagination, adding `page=x`.

**Note** - macro will not need pagination. There are only 33 rows possible, so we should just show them all.

wpears:
Closes #274, remaining paginated components will have individual associated tickets

Implements the structure of pagination as well as the concrete implementation for parse errors.

### DIFF
--- a/src/js/components/Pagination.jsx
+++ b/src/js/components/Pagination.jsx
@@ -21,4 +21,11 @@ const Pagination = (props) => {
   )
 }
 
+Pagination.propTypes = {
+  pagination: PropTypes.object,
+  getPage: PropTypes.func,
+  getPreviousPage: PropTypes.func,
+  getNextPage: PropTypes.func
+}
+
 export default Pagination
--- a/src/js/components/Pagination.jsx
+++ b/src/js/components/Pagination.jsx
@@ -1,0 +1,24 @@
+import React, { Component, PropTypes } from 'react'
+
+const Pagination = (props) => {
+  const page = props.pagination
+  if(!page) return null
+  const firstPage = page._links.self === page._links.first
+  const lastPage = page._links.self === page._links.last
+
+  return (
+    <div className="PaginationControls">
+      <button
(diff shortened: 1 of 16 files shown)


src/js/components/Pagination.jsx
```diff
@@ -0,0 +1,24 @@
...
+      <div>Page <h4>{page._links.self.slice(-1)}</h4> of {Math.ceil(page.total/20)}</div>
+      <button
+        className={ lastPage ? 'usa-button-disabled' : '' }
+        onClick={ e => { if(!lastPage) props.getNextPage(page) }}
+      >Next</button>
+    </div>
+  )
+}
+
+export default Pagination
```
awolfe76: Can we add defaultProps here?

wpears: I've turned away from using defaultProps for components that are always wrapped by a container because I think it confuses what the source of truth for the props is (which, imo, should always be the state in the redux store).

If you wanted them there to show what the component accepts, I'd say we could just put in some `PropTypes`

awolfe76: Yeah, sorry 😞 . I meant PropTypes. I find it useful just to see what should be coming in.

awolfe76 (commented):
Awesome! 🎆

awolfe76:
@wpears - I'll merge when Travis has finished. (If you notice it first go ahead.)

Issue #402, opened 2017-03-02
Update edits nav to show progress


awolfe76:
The edits nav doesn't show any idea of it being "steps" to go through to complete a submission.

As an example, https://dribbble.com/shots/1711346-Simpler-Checkout-Process.

PR #430, opened 2017-03-13
Edits nav

awolfe76:
Does part of #402. There is some code in place in `sass/components/EditsNav.scss` and `js/components/EditsNav.jsx` (commented out) to handle the rest but I think this will be much simpler when https://github.com/cfpb/hmda-platform/issues/849 is complete.

For now this PR:
- cleans us the header a bit
- improves edits navigation and user heading

#### Screenshot
updated-nav

### DIFF
--- a/__tests__/components/EditsNav.js
+++ b/__tests__/components/EditsNav.js
@@ -1,8 +1,9 @@
 jest.unmock('../../src/js/components/EditsNav.jsx')
 
-import EditsNav,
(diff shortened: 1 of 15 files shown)


wpears:
~~I'm seeing an error `Cannot read property 'edits' of undefined` when trying to switch between the items in the the nav. Investigating.~~

Was a backend error, fixing errors (ie #431) will prevent these kind of syntax-ish errors in the future.

Looks good :+1:

PR #433, opened 2017-03-16
ErrorWarning component implemented and error key added to state

### LINKED ISSUE #431: Error notification component
Errors everywhere should update an `error` key in the state which then renders a warning.
Other component's handlers should probably be disabled (and all buttons moved to disabled state) when an error exists.

wpears:
Closes #431 

Picks out 403, 404, and 500 for special error messages. These can be adjusted as needed.

An error existing only prevents signing, if desired this could be extended to verification in a similar manner.

### DIFF
--- a/src/js/api.js
+++ b/src/js/api.js
@@ -56,6 +56,7 @@ export function sendFetch(options = {method: 'GET'}){
         return response.text()
       }
       const json = response.json()
+      console.log('got api response for', url, json)
       return json
     })
 }
--- a/src/js/components/ErrorWarning.jsx
+++ b/src/js/components/ErrorWarning.jsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { PropTypes } from 'react'
 
 export function renderHeader(props) {
   if(props.headerText) return props.headerText
@@ -47,6 +47,12 @@ const ErrorWarning = (props) => {
       </div>
     </div>
   )
+}
+
+ErrorWarning.propTypes = {
+  error: PropTypes.object,
+  bodyText: PropTypes.string,
+  headerText: PropTypes.string
 }
 
 export default ErrorWarning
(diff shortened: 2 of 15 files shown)


src/js/api.js
```diff
@@ -56,6 +56,7 @@ export function sendFetch(options = {method: 'GET'}){
         return response.text()
       }
       const json = response.json()
+      console.log('got api response for', url, json)
```
awolfe76, line 59: Can probably remove this logging.

src/js/components/Signature.jsx
```diff
@@ -14,8 +14,19 @@ const showReceipt = (code, timestamp, receipt) => {
   )
 }
 
-const showWarning = (code) => {
-  if(code > 8) return null
+const showWarning = (props) => {
+  if(!props.error && props.status.code > 8) return null
+
+  if(props.error) {
+    return (
```
awolfe76: Why not use the error component here? If it's for the specific message maybe another `prop` would help? My initial take is that it would be nice to have this done in one place.

wpears: :+1:

src/js/api.js
wpears: Can we keep this in? I think it pairs well with the log above it without being too spammy

awolfe76: Yeah, ok with me.

src/js/components/ErrorWarning.jsx
```diff
@@ -0,0 +1,52 @@
...
+    <div className="ErrorWarning usa-alert usa-alert-error">
+      <div className="usa-alert-body">
+        <h3 className="usa-alert-heading">{renderHeader(props)}</h3>
+        <p className="usa-alert-text">{renderBody(props)}</p>
+      </div>
+    </div>
+  )
+}
+
+export default ErrorWarning
```
awolfe76: Can we add `propTypes` here too?

awolfe76 (approved):
👍

PR #440, opened 2017-03-17
Column width and header update to institutions

### LINKED ISSUE #439: Content and styling for Institutions page
I think we need a header and subheader about the page, something like

### Filing period 2017

Also, I think that for visual balance we should change the two-thirds/one-third division of the list of institutions and the explanatory texts to one-half/one-half.

wpears:
Closes #439


awolfe76 (approved):
I like it. I think we can play around with the headings across to the app to make them consistent but this sets it up. 👍

Issue #432, opened 2017-03-13
Don't collapse columns during data fetch


wpears:
Along with https://github.com/cfpb/hmda-platform-ui/issues/322, during a data fetch, we need to ensure that the layout elements of the application are static / not based on presence/absence of data (as there will always be no data before the first fetch). Action items:
 - Ensure we're appropriately using `us-` widths on all components
 - Return empty/placeholder divs rather than null when no data is present (the placeholder may be eg, the spinning loading icon)
 - Ensure that validity of the components and their tests doesn't rely on the presence/absence of a rendered component (there will be updates here where tests are checking for `null` on components rendering with no data.

awolfe76:
Closed by #445

PR #445, opened 2017-03-20
Various visual bugs

wpears:
Don't be fooled by the branch name, I wasn't actually able to repro the column collapsing bug, so I didn't fix it.

Coolest fix is removing an extraneous call to the submissions endpoint


awolfe76 (approved):
👍

wpears:
Closes #432

PR #455, opened 2017-03-30
Get Env vars from json file

### LINKED ISSUE #452: Ennvars from external source, held in redux state
This allows us to do the JS build at docker build time without having to do any (potentially) tricky find and replace stuff at container startup.

This *also* provides a mechanism for these values to change over time, which I don't think should happen due to how the proxying work, but it's probably worth the slight delay to first page load to have this flexibility.

wpears:
Rather than baking in env vars at JS build time, this PR changes various build process to expect them at runtime in a json file located at `/env.json` which gets fetched and loaded at app bootstrapping. The docker `RUN` command first runs a small shell script that builds this file before starting nginx.

The main thing this saves is having to run the js build at container startup (as that is when the values of the needed environment variables are determined). The cost is a small delay on first page load.

Closes #452


Dockerfile
```diff
@@ -1,14 +1,20 @@
 FROM nginx:1.10
 MAINTAINER Wyatt Pearsall<Wyatt.Pearsall@cfpb.gov>
 
-RUN mkdir -p /usr/src/app
+RUN apt-get update && \
```
awolfe76: Do we need to bring this all back?

wpears: If we want only one Dockerfile, then yeah. We can bring it up in discussion today.

hkeeler: I think the short answer for now is _yes_.  We _could_ have a separate `Dockerfile.nginxonly` or perhaps get clever with buildargs, but I'd rather keep the setup simple for now.

Dockerfile
```diff
@@ -1,14 +1,18 @@
 FROM nginx:1.10
 MAINTAINER Wyatt Pearsall<Wyatt.Pearsall@cfpb.gov>
 
-RUN mkdir -p /usr/src/app
+RUN if [ -z ${SKIP_JS_BUILD+x} ]; then echo "Installing JS deps" && apt-get update && \
+    apt-get install -y curl g++ git make && \
```
awolfe76: We can remove `git` from here. That was required because of USWDS and the latest release from them fixed it.

awolfe76:
One last thought here ... what if we moved our approach closer to what the back-end does? There they don't build the `.jar` within the [`Dockerfile`](https://github.com/cfpb/hmda-platform/blob/master/Dockerfile), its already built. The CI job handles running `sbt clean assembly`.

Why wouldn't we do the same thing, at least to be consistent? We'd run our `npm run build` prior (like we do currently for the dev setup) and have the CI job do the same. It would definitely shrink our image and start-up time.

bootstrap.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
hkeeler: Should we name this `docker-entrypoing.sh`?  I think that's the semi-standard.

Also, to make sure `./env.sh` is successful, shouldn't we do:

```bash
./env.sh && nginx -g "daemon off;"
```

env.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
```
hkeeler, line 1: How about a `set -e` so the script will exit if any part fails?

env.sh
```diff
@@ -0,0 +1,24 @@
...
+envvars=( 'APP_URL' 'HMDA_API' 'KEYCLOAK_URL' )
+defaults=( 'http://192.168.99.100' 'https://192.168.99.100:4443/hmda' 'https://192.168.99.100:8443/auth/realms/hmda' )
```
hkeeler: I know I brought up adding defaults when we talked about this, but do we really want them?  What will the experience be the defaults are applied, but they're wrong?  Will it be apparent to the dev what's going wrong?  Also, are you planning on setting them in `docker-compose.yml`?  Seems like you should just to give the hint of what needs to be changed when you're not using the default...and if it is going to be there, why even have the default?

In comparison, [`keycloak/docker-entrypoint.sh`](https://github.com/cfpb/hmda-platform-auth/blob/master/keycloak/docker-entrypoint.sh) exits on almost all of the envvars if they're not present.

env.sh
hkeeler, line 1: This script is pretty clever, but it seems a little complex for setting 3 values in a config file.  Could we not just have a template `env.json` with placeholders, and do a `sed` find/replace like we do in [`keycloak/docker-entrypoint.sh`](https://github.com/cfpb/hmda-platform-auth/blob/master/keycloak/docker-entrypoint.sh).

If you do want to keep it this way, I think it at least deserves some comments describing the intent of the script.

Dockerfile
```diff
@@ -1,14 +1,18 @@
 FROM nginx:1.10
 MAINTAINER Wyatt Pearsall<Wyatt.Pearsall@cfpb.gov>
 
-RUN mkdir -p /usr/src/app
+RUN if [ -z ${SKIP_JS_BUILD+x} ]; then echo "Installing JS deps" && apt-get update && \
```
hkeeler: Where does `SKIP_JS_BUILD` come from?  Is that an `ARG`-yet-to-come?

hkeeler (commented):
I like it!  I added a few comment.  Should this be a _WIP_?  Seems like there are some bits that aren't yet included, mostly related to `SKIP_JS_BUILD`.

env.sh
awolfe76: +1 to the template idea.

env.sh
wpears: The reason the defaults are included is so devs working from the volumized compose setup don't have to set these envvars on their machines.

env.sh
wpears: I'll change it over.

wpears:
So the reason I never brought up the local js build is twofold:
 -  I don't really *like* that you have to generate the jar in the backend before bringing up the the compose setup
 - I wasn't certain about how much we wanted to depend on Jenkins (ie in this case Jenkins would have to do the front-end build and so have all the front-end deps at the right versions).

I *like* the idea, ~~especially cause the `build-args` thing is a bit fraught (we can't use `build-args` with the current version of docker-machine we're running~~ )

Also, this would mean we're requiring anyone who wants to setup the full platform locally to have node/npm rather than just docker (which, tbh, if we require them to have sbt, we're already kinda relegating this to dev machines).

The real solution here, of course, either using the cluster for stuff non-devs will want to use the app for, or actually pushing up the docker images built by the devs to be reused without any deps other than docker.

Dockerfile
wpears: SKIP_JS_BUILD would be provided as an `arg` in hmda-platform's compose file.

~~However this approach might not be the right course atm due to the version of docker machine we're stuck on~~

hkeeler:
Per @awolfe76 
>One last thought here ... what if we moved our approach closer to what the back-end does? There they don't build the `.jar` within the [`Dockerfile`](https://github.com/cfpb/hmda-platform/blob/master/Dockerfile), its already built. The CI job handles running `sbt clean assembly`.
>
>Why wouldn't we do the same thing, at least to be consistent? We'd run our `npm run build` prior (like we do currently for the dev setup) and have the CI job do the same. It would definitely shrink our image and start-up time.

It's a good question.  It is a tradeoff of fewer steps and tools vs. slimmer, faster-building images.   Personally, I'd prefer if `hmda-platform` did work this way...and if this style works out here, we should consider doing it there too.  The more steps we have outside of the Docker build, the farther away we get from having `docker-compose up` just work, and the general expectation that `docker build` does everything you need.  It's also more steps to configure and keep in-sync in CI.

If we were to move all images to this pattern, you'd have to do something like this to build everything:

```bash
sbt clean assembly
cd ../hmda-platform-ui
npm run build
cd ../hmda-platform-auth/keycloak/providers/authenticator/hmda/
mvn clean install
cd ../../../../../hmda-platform
docker-compose build ...
```
...and who know what other services we'll cook up.  The more we add, the more complex the build gets.
I like just `docker-compose up --build` a lot better.  This also means you now need `sbt`, `npm`, and `mvn` locally just to launch the stack, and making sure whoever launches it has the right version of those tool, etc., etc...

I really think this pattern could be the best of both world.  With a build arg, we can have these images default to doing _everything_, but switch to a dev-friendly, fast-building image with just a flag.

hkeeler:
Per @wpears:
>I *like* the idea, especially cause the `build-args` thing is a bit fraught (we can't use `build-args` with the current version of docker-machine we're running 😭 😭 )

Which version of Docker/Compose do we need for this to work?

awolfe76:
I think we should update the platform dockerfile to do this too then. In the end my point is just that we're consistent, which we're not right now. I agree that one simple command to start the whole thing is ideal but on the dev side, because of mounted volumes, we're likely to always need the extra tools to make things smooth.

wpears:
>Which version of Docker/Compose do we need for this to work?

Oh. In responding to this I realized that what I said before kinda missed the mark, as we're declaring compose file version 2, which *should* [support build-args](https://docs.docker.com/compose/compose-file/compose-file-v2/)

awolfe76:
Created https://github.com/cfpb/hmda-platform/issues/907 to update the platform dockerfile to include `sbt`.

wpears:
In the setup proposed in https://github.com/cfpb/hmda-platform/pull/910, the JS build will be skipped by default. We could create another docker-compose without the build arg or volume for the full build to be included in the image.

Whether we want to have completely separate dockerfiles or build-arg dependent paths in a single dockerfile is a point up for discussion.

awolfe76:
@wpears - I get a blank page in the UI and see `Error fetching connection information SyntaxError: Unexpected token < in JSON at position 0`. Looks like its not getting the `env.json` correctly?

wpears:
@awolfe76 thanks, `./dist` got dropped in the switch over to `sed`ing the template

awolfe76 (approved):
@wpears - I'll let you merge along with https://github.com/cfpb/hmda-platform/pull/910 so we get them both at the same time.

hkeeler:
I'm doing some local testing on all of this now.  If it all checks out, I'll merge both PRs.

Dockerfile
```diff
@@ -1,14 +1,19 @@
...
+    apt-get install -y nodejs; fi && mkdir -p /usr/src/app
 
 WORKDIR /usr/src/app
-# COPY . /usr/src/app
+
+COPY . /usr/src/app
 COPY nginx/nginx.conf /etc/nginx/nginx.conf
 
-# RUN chown -R root /usr/src/app && chmod u+rwx /usr/src/app
+RUN if [ -z ${SKIP_JS_BUILD+x} ]; then echo "Building JS" && npm install; fi
```
hkeeler: Whenever I build this without `SKIP_JS_BUILD` set, and I run it without the dev volume mounted, I end up with a near-empty `dist` directory.  To fix this, I had to add `&& npm run build` to this line.  Is see there is a [`postinstall`](https://github.com/cfpb/hmda-platform-ui/pull/455/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R19) script, that seems to do this, and I assume run _post-`install`_, but I don't think its workin'...if that's the intent.

wpears: Good catch, npm was failing do to preferring to not run as root. This was actually quite a nice time to switch to running yarn in the dockerfile, so I switched it over (installs in most cases should proceed from the local copies of modules).

You can safely ignore the changes to the various yarn files.

hkeeler:
I can confirm that this does now work as intended, both in dev-mode with the buildarg `SKIP_JS_BUILD` and the mounted volume, **and** the with-all-dependencies build (no `SKIP_JS_BUILD` or volume).

Now, whether adding a whole new package manager is appropriate for this PR, and whether we actually want to be committing dependencies into our repo...I dunno.  In my past experience, that's generally not a great idea, but perhaps that's how yarn is supposed to work?

@awolfe76, thoughts?  If you're OK with this, I guess I am too.

wpears:
We actually added yarn to the project a few weeks ago but never moved it to the docker build. Tbh I was pretty squeamish about it as well, but it has several nice properties.

Also, when these dependencies DO change, it just bumps the tarball, which essentially requires no review in github (rather than if we'd included `node_modules`, where including the new dep would show up as normal code changes)

PR #483, opened 2017-04-11
Upload bugs

### LINKED ISSUE #470: Upload page temporarily freezes after pressing "Upload"
When I uploaded a large file (`clean_14117-lars.txt`), I saw about a 4 second lag between pressing the "upload" button and seeing the gray/green dots that show the status of "uploading", "parsing", "validating".

Not sure if it's avoidable, but I wanted to let you know.

### LINKED ISSUE #477: Cancel polling on selecting a file

wpears:
Closes #477 
Closes #470

### DIFF
--- a/src/js/actions/index.js
+++ b/src/js/actions/index.js
@@ -481,8 +481,9 @@ export function fetchCSV(institutionId, filing, submissionId) {
 /*
  * Wire upload together with xhr so progress can be tracked
  */
-export function requestUpload(file) {
+export function fetchUpload(file) {
   return dispatch => {
+    dispatch(uploadStart())
     var data = new FormData()
     data.append('file', file)
 
@@ -511,7 +512,6 @@ export function requestUpload(file) {
     xhr.setRequestHeader('Accept', 'application/json');
     xhr.send(data);
 
-    dispatch(uploadStart())
     return Promise.resolve()
   }
 }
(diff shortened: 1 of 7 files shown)


src/js/reducers/index.js
```diff
@@ -18,6 +18,7 @@ import {
   SHOW_CONFIRM,
   HIDE_CONFIRM,
   PICK_SORT,
+  UPLOAD_START,
```
awolfe76, line 21: Shouldn't this require a change to `constants` too?

src/js/actions/index.js
```diff
@@ -26,6 +26,7 @@ import fileSaver from 'file-saver'
 
 let latestSubmissionId
 let currentFilingPeriod
+const pollObj = {polling: false}
```
awolfe76, line 29: Why an object here?

src/js/reducers/index.js
wpears: It was already in constants, just wasn't being mapped to the state.

src/js/actions/index.js
wpears: So that `pollForProgress can be passed a value that can be altered externally. Using an object means you can:
 - in the tests for `pollForProgress` pass in the value to check the behavior for each condition
 - change the polling value externally (ie from `setFile` or `setNewFile`

If it was just a boolean, you couldn't meet both these conditions simultaneously

awolfe76: Thanks!

awolfe76 (approved):
👍

Issue #468, opened 2017-04-07
Visual separation of edit tables


wpears:
To assist with #466 (or perhaps to address the root cause), edit tables should be visually separated more.

This could include any combination of:
 - `<hr/>`s
 - More whitespice
 - Larger headers
 - Alternating background colors (between white and `$lightest-gray`) for each `EditsTableEntry`

awolfe76:
I've had a few thoughts about this.

- white space is a great idea, and maybe when combined with the others that you mention, or
- making them more "card" like ([example](https://dribbble.com/shots/1790652-Google-Material-exploration), base on [this](https://material.io/guidelines/components/cards.html)), so having the `$lightest-gray` as the background and the edits with a white bg


I think we can play and demo here a bit.

PR #490, opened 2017-04-12
Don't render tables for macro

### LINKED ISSUE #489: Update macro edits to not render a table
Remove rendering the tables for macro edits. Having the fact that the edit was triggered along with its description is enough to allow the filer to verify the edit.

_See https://github.com/cfpb/hmda-platform/issues/908 for more._

wpears:
Closes #489 

The separation of edits in #468 will be especially important for macro now (as such I've essentially ignored the styling of the macro page in this PR)

### DIFF
--- a/src/js/components/EditsTable.jsx
+++ b/src/js/components/EditsTable.jsx
@@ -19,8 +19,6 @@ export const renderHeader = (edits, rows, type) => {
   let keyCells = rows[0].row
   const fieldCells = rows[0].fields
 
-  if(type === 'macro') keyCells = {}
-
   const numOfCells = Object.keys(keyCells).length + Object.keys(fieldCells).length
   const cellWidth = `${100/numOfCells}%`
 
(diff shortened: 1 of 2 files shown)


src/js/components/EditsTable.jsx
```diff
@@ -52,8 +50,17 @@ export const renderTableCaption = (edit, rowObj, pagination) => {
...
+      <div className="caption">
+        <h3>{captionHeader}</h3>
+        {description ? <p className="usa-font-lead">{description}</p>:null}
+      </div>
+    )
+  }
+
   return (
-    <caption>
+    <caption className="caption">
```
awolfe76, line 63: Is the `className` needed here? (And [above](https://github.com/cfpb/hmda-platform-ui/pull/490/commits/aa6ec7be754e477479a18f552da12af63ef1d465#diff-d1bd5067e44fa2ba9bdd347727b4fd51R55))

wpears: Not strictly needed, but I felt like it would be wise to include a way to target the caption whether it was a true caption or just a div.

awolfe76: Ok, I'll merge because I'd like to have this as part of my pagination updates but would rather not add unused classes. It can be a bit misleading when looking at/for the CSS.

PR #494, opened 2017-04-18
Edits nav edits independence

### LINKED ISSUE #476: Edit nav rendering incorrectly if edits haven't been fetched
If you visit the summary page first, which could happen from the institution's page based on the status of the current submission, the nav will render incorrectly.

edits-nav

This is because it's partially based on the edits.

https://github.com/cfpb/hmda-platform/issues/849 should help us fix that by being able to remove the dependency on edits and making it completely based on the submission status, but wanted to create this issue to make sure we come back to it.

wpears:
Closes #476 

Also, routes away from pages that shouldn't be viewable based on app state (eg, the summary page before all edits are verified or macro edits when synval exist). These weren't linked in the app at the wrong times, but still could be visited via URL (and could have possibly led to confusing situations if users attempted to bookmark pages across submissions).

### DIFF
--- a/__tests__/components/EditsTableWrapper.js
+++ b/__tests__/components/EditsTableWrapper.js
@@ -19,10 +19,10 @@ const types = {
 describe('EditsTableWrapper', () => {
   const onDownloadClick = jest.fn()
 
-  it('renders 4 children if type is quality and typeFromPath is quality', () => {
+  it('renders 4 children if type is quality and page is quality', () => {
     const rendered = EditsTableWrapper({
       onDownloadClick: onDownloadClick,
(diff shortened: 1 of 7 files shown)


awolfe76:
@wpears - one thing I've noticed is that if you are on the "summary" page and hit refresh (from #476) the nav is close to correct ... the "Summary" link is active but the progress number is not. It looks like this:
summary-not-active
It should have an `active` class on the `<li>`.

But, the bigger issue is that when I hit refresh from the "summary" page I'm getting pushed to the "syntacticalvalidity" page.

awolfe76 (approved):
👍

Issue #471, opened 2017-04-07
Fixed submission navigation


awolfe76:
The navigation through the submission process should remain in view at all times. This can work a few different ways. 

If we want the header to remain fixed too, its all pretty straightforward.

If we do NOT want the header to remain fixed we'll have to detect scrolling and position the submission navigation as fixed when it reaches the top of the window.

awolfe76:
We could also have the submission navigation be fixed as a side navigation ([USWDS has this](https://standards.usa.gov/components/sidenav/) and we could customize it to still show the idea of steps/progress.)

PR #499, opened 2017-04-21
Home page and nav updates

### LINKED ISSUE #487: Home page and main navigation design
I think we're at a point now that we can re-work the home page structure to make it a little more friendly. Some possibilities:

- add a banner to highlight registering and login
- maybe don't show the institutions link until logged in
  - also show it next to the user name
- move the logo to the footer, and make the "title" be something like "HMDA filing"

awolfe76:
Closes #487 

- adds a "hero" section to the home page to bring a focus to getting logged in
- always redirects to the institutions page after login
  - this seems to be a more useful location to end up (the home page isn't very useful when the goal is filing
- removes navigation for institutions until logged in as well, when logged in this link is highlighted more by being a button (outlined button)
- moves logo to footer and replaces it with "HMDA filing" in the header

#### New home page (no real navigation, but we could add an FAQ page)

new-home

---

#### Navigation after login

new-nav

---

#### New home page after login (to show that it still exists, we can/should update this content too)

new-home-logged-in

### DIFF
--- a/__tests__/components/Header.js
+++ b/__tests__/components/Header.js
@@ -1,7 +1,7 @@
 jest.unmock('../../src/js/components/Header.jsx')
 jest.mock('oidc-client')
 
-import Header from '../../src/js/components/Header.jsx'
+import Header, { addActiveClass } from '../../src/js/components/Header.jsx'
 import Wrapper from '../Wrapper.js'
 import React from 'react'
 import ReactDOM from 'react-dom'
@@ -22,8 +22,8 @@ describe('Header', () => {
     expect(headerNode).toBeDefined()
   })
 
-  it('render the logout link', () => {
-    expect(TestUtils.scryRenderedDOMComponentsWithClass(header, 'logout').length).toBe(1)
+  it('render the institutions link', () => {
+    expect(TestUtils.scryRenderedDOMComponentsWithClass(header, 'nav-institutions').length).toBe(1)
   })
 
   const headerNoUser = TestUtils.renderIntoDocument(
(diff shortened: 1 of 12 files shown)


wpears:
Looks cool, a couple notes:

The sticky header is inconsistent between the homepage, the institution page, and the submission page (doesn't stick on homepage, doesn't fully respect the header on the institutions page, works as expected on the submissions page). I think this is due to the top margin being set on the element after the header varying between each page (for example, the `SubmissionContainer` gets a `marginTop` of `4em`.

I like the presence of the `MY INSTITUTIONS` button but I think we should dial back the visual weight (right now it almost looks like it's always active, as if you're always on the page). I think a more consistent design would be to add an underline to both the `home` and `my institutions` links when they are active, similar to how it works for upload, synval, etc in the submission process.

I like the re-org of the logo and the HMDA Filing text. Might be worth thinking of adding a more interesting font for the header that could be used for all satellite HMDA products.

wpears (changes requested):
(see inline comment)

awolfe76:
@wpears 

> The sticky header is inconsistent between the homepage, the institution page, and the submission page (doesn't stick on homepage, doesn't fully respect the header on the institutions page, works as expected on the submissions page). I think this is due to the top margin being set on the element after the header varying between each page (for example, the SubmissionContainer gets a marginTop of 4em.

There shouldn't be a sticky header at all. #471 might do something there but this PR shouldn't be.

> I like the presence of the MY INSTITUTIONS button but I think we should dial back the visual weight (right now it almost looks like it's always active, as if you're always on the page). I think a more consistent design would be to add an underline to both the home and my institutions links when they are active, similar to how it works for upload, synval, etc in the submission process.

I changed things up a bit here. It now looks like:
nav
You'll see the home page link is gone. I think this is ok (at least for now) because the title is a link to home, which is a pretty common way of handling this. 

> I like the re-org of the logo and the HMDA Filing text. Might be worth thinking of adding a more interesting font for the header that could be used for all satellite HMDA products.

Yeah, I think we can look into this in a different issue/PR.

wpears (approved):

lesgou:
late to the game, but looks cleaner and less busy

PR #510, opened 2017-04-26
Increase coverage

wpears:
Reorganizes some stray files and pulls several small modules out of `api.js` for testing.


__tests__/api/api.js
```diff
@@ -0,0 +1,10 @@
+jest.unmock('../../src/js/api/api')
+
+import * as api from '../../src/js/api/api'
+
+describe('api', () => {
+
+  it('fetches institutions', () => {
```
awolfe76: More tests coming?

awolfe76 (changes requested):
I like what you did with the `utils`, but are there more tests coming in a few places? The `api`, `fetch`, and `actions` seem be :hurtrealbad: .

wpears:
@awolfe76 should be good to go now

awolfe76 (approved):

PR #513, opened 2017-04-27
fix nav breaking down, now at 963px

### LINKED ISSUE #469: Funky Alignment in Navigation Bar
4 of the 5 circles are below the line, 1 is on it. 
(Ignore the "Automatically logged out" message in the screenshot; the same alignment issue is true on all submission pages)

screen shot 2017-04-07 at 3 31 04 pm

awolfe76:
Closes #469 ... again. 😄

### DIFF
--- a/src/sass/components/EditsNav.scss
+++ b/src/sass/components/EditsNav.scss
@@ -3,18 +3,11 @@
   text-align: center;
   ul.usa-nav-primary {
     li {
-      width: 20%;
       text-align: center;
       margin-bottom: 2em;
-
-      // the edits nav breaks down at 963 and below because the "syntactical and validity edits" gets broken into 2 lines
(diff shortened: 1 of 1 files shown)


wpears (changes requested):
This still seems to break under 963px and continues to degrade at smaller sizes. I'd say if we're aiming to fix realignment problems at various breakpoints we should have a solution for all sizes (ie, redoing the CSS for the alignment of the numbered circles so they are completely independent of line breaks in the associated text.

awolfe76:
@wpears - check out 563f74b470f9d83ff9f76e12d8f67b5e43df119c. Much cleaner and seems to do the trick.

wpears:
Bingo! :1st_place_medal:

PR #537, opened 2017-05-12
Split actions

### LINKED ISSUE #503: Split actions and reducers into several files
Both of these files are getting difficult to work with and can be split into several files in order to manage them easier.

wpears:
Partially closes #503 

Separate modules created for all actions and action tests (with missing tests mocked for now)


awolfe76 (approved):
👍  - looks good but there are a few conflicts now.

PR #539, opened 2017-05-13
Poll for summary, with cancel on unmount

### LINKED ISSUE #255: Poll for when the IRS report is generated
~~Depends on https://github.com/cfpb/hmda-platform/issues/647.~~

~~When the status code changes to `9: validated` we should start polling the `/latest` endpoint to watch for the status to change to `10: IRS report generated` and then make the request for the report.~~

~~Right now we have no way of knowing when we get to `10`.~~

Depends on https://github.com/cfpb/hmda-platform/issues/767. Check there for more details.

We should start polling for the IRS report when the summary "page" is loaded AND status is > `7: validating`. This is because the IRS report has no dependencies on edits.

The component should render an "information" message if the criteria above isn't met.

We should also stop polling if the IRS report isn't ready and the filer leaves the summary "page".

wpears:
Closes #255 

`fetchSummary` didn't have tests to start with and this will conflict with #537, so I didn't include extra tests here (it's likely that the architecture will change somewhat with #537 merged).

But anyway, this issue has been annoying me, so here's the fix :v:


awolfe76 (changes requested):
I think #255 was to do this for IRS, which I think it more likely to need it. The IRS report could take a while to generate for larger filers, where the summary information would almost immediately be available.

wpears:
Hah! I suppose I should stop conflating summary and IRS in my mind.

Geez louise. Fixed.

awolfe76 (approved):

PR #540, opened 2017-05-16
Split reducers

### LINKED ISSUE #503: Split actions and reducers into several files

wpears:
Closes #503
(with #537)
Splits reducers into separate files


awolfe76 (approved):
👍

PR #544, opened 2017-05-17
Auto-redirect from fetch

### LINKED ISSUE #538: Need to bubble auth errors up/have clear messaging upon signout
The token should be refreshed when possible, but if it isn't, the user to should redirected to a "you have been signed out" page (or show an error saying the same)

wpears:
Closes #538 

If met with 401s (SSO token expired, unable to silent renew), any failed request will redirect the oidc-callback page to refresh the token. This will fail if the session expiration has been reached (but this defaults to 8 hrs, so not something we should be strictly concerned about).


awolfe76 (approved):
👍

PR #551, opened 2017-05-19
Present clear message to unsupported browsers

### LINKED ISSUE #542: Only show browser text on detecting an old browser / make it blocking
This could go as far as replacing the hero with an error type message:

We're sorry, in order to securely file your data you will need to use a modern browser.
Please try again with Google Chrome, Mozilla Firefox, Internet Explorer 11, Microsoft Edge, or Safari.

### LINKED ISSUE #550: User Header isn't rendering properly on homepage
Refreshing the page on the homepage after logging in (when the logged in header and the logged in hero are shown) will result in these components being set to the unlogged in state, even though the user is still logged in.

wpears:
Also, bonus reorg of user heading
Closes #542


awolfe76:
When running this locally I can't get to the home page. I'm logged out and `http://192.168.99.100/` redirects me to the login page instead of the app home page.

And, once I'm logged in I can't seem to log back out.

awolfe76 (approved):
👍

Issue #547, opened 2017-05-17
Accessibility testing


awolfe76:
Come up with a way to do both automated and manual accessibility testing. It would be nice if it worked with our auth setup.

Possible tools:
- [wcag](https://github.com/cfpb/node-wcag) (I've tried and can't get it to run)
- [pa11y](https://github.com/pa11y/pa11y) (this one does run, but haven't tried to run as authenticated ... can test the home page but other pages render empty)
  - `pa11y -r csv -S ./__tests__/file.jpg http://192.168.99.100 > report.csv` works
  - `pa11y -r csv -S ./__tests__/file.jpg http://192.168.99.100/6999998/2017/macro > report.csv` also works but the screen shot (`-S`) is empty

_This issue will likely create other issues to fix any accessibility problems._

awolfe76:
We've [done some manual testing](https://github.com/cfpb/hmda-platform-ui/pull/556) (should do more) and it would be good to have something automated as well. Leaving this open and carrying over.

awolfe76:
Some findings:

- [x] `/institutions` has an info alert, the link inside of it doesn't have sufficient contrast
- [x] `/submission` has 2 `signature` ids, the `<section>` and the `<input>` (checkbox)
- [x] in the edits navigation, the grayed out version of the links do not have sufficient contrast
- [x] when edits exist on `/syntacticalvalidty` the "download the edit report" link doesn't have sufficient contrast
- [ ] when quality and macro edits exist, the orange nav doesn't have sufficient contrast
- [x] because pagination is a form and has an input it requires a label
- [x] the grayed out button for pagination doesn't have sufficient contrast
  - this will be fixed by #811 

_Found using https://www.deque.com/products/axe/._

awolfe76:
Also see https://github.com/cfpb/hmda-platform-auth/issues/169.

PR #556, opened 2017-05-23
Accessibility updates

awolfe76:
This PR is the first step toward #547, there will be at least one more.


wpears (approved):
Cool. The changes make sense. We'll need to watch ids on the IRS page or think about prefixing the ids (ie irs-loan-type, irs-conv, etc).

PR #561, opened 2017-05-25
Protect uppercase operation if agency missing

wpears:


awolfe76 (commented):
Not that this is wrong, but would it be easier to just add [this class](https://github.com/cfpb/hmda-platform-ui/blob/master/src/sass/hmda.scss#L48) to it?

awolfe76 (approved):

PR #580, opened 2017-05-30
Add status code constants

### LINKED ISSUE #565: Create a properties file for submission statuses
The submission statuses are used throughout the front-end to determine what to render and when. Its happened a few times that the statuses have been updated and some of the locations have been missed.

It would be helpful to create a file with these constants to read in and then use.

[Current statuses](https://github.com/cfpb/hmda-platform/blob/48c9a76759816227d7de8857be1e0ccf1abffa29/Documents/submission-status.md)

Examples (there are plenty of others) where this can be used ... [this](https://github.com/cfpb/hmda-platform-ui/blob/1096c33730a8174d5ddbed097e881cec46bc215e/src/js/containers/Submission.jsx#L32) and [this](https://github.com/cfpb/hmda-platform-ui/blob/1096c33730a8174d5ddbed097e881cec46bc215e/src/js/components/EditsNav.jsx#L37).

wpears:
Closes #565

### DIFF
--- a/__tests__/components/UploadForm.js
+++ b/__tests__/components/UploadForm.js
@@ -64,6 +64,7 @@ describe('submitform', function(){
           setFile={setFile}
           uploading={true}
           file={{size:200}}
+          errors={[]}
         />
       </Wrapper>, node)
   const form2Node = ReactDOM.findDOMNode(form2)
--- a/__tests__/reducers/signature.js
+++ b/__tests__/reducers/signature.js
@@ -7,10 +7,6 @@ const defaultSignature = {
   isFetching: false,
   timestamp: null,
   receipt: null,
-  status: {
-    code: 0,
-    message: ''
-  },
   checked: false
 }
 
--- a/src/js/actions/pollForProgress.js
+++ b/src/js/actions/pollForProgress.js
@@ -3,6 +3,7 @@ import receiveSubmission from './receiveSubmission.js'
 import receiveError from './receiveError.js'
 import hasHttpError from './hasHttpError.js'
 import { getLatestSubmission } from '../api/api.js'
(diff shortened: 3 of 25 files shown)


awolfe76 (changes requested):
We should probably use the new constants in the reducers as well, eg [here](https://github.com/cfpb/hmda-platform-ui/blob/b9b0474b17e5a4283374eb68c27a4d458e545811/src/js/reducers/signature.js#L8) and [here](https://github.com/cfpb/hmda-platform-ui/blob/b9b0474b17e5a4283374eb68c27a4d458e545811/src/js/reducers/submission.js#L6).

Also having an `UNINITIALIZED = 0` feels a bit strange to me since it doesn't exist on the back-end. It looks like we're only ever doing `if(!this.props.status || this.props.status.code === UNINITIALIZED)` with that status. Could we maybe handle that differently and remove the `0`?

I also noticed [this](https://github.com/cfpb/hmda-platform-ui/blob/b9b0474b17e5a4283374eb68c27a4d458e545811/src/js/components/ConfirmationModal.jsx#L56) from the ConfirmationModal, which we can probably just remove completely. _(For the `defaultProps`, we still have these in a few components, editsTableCell, Institutions, Signature, and UploadForm. We can create an issue to clean them up or get rid of them here.)_

wpears:
I think `UNINITIALIZED` not existing in the backend is precisely what makes it useful ^_~ since we know that we're before the initial fetch and can act appropriately. Another approach would be to default `code` to `null`, but then that muddies our `PropTypes` a bit.

awolfe76:
`null` is what I was thinking but wasn't thinking about the `PropTypes`. I think your approach is :1st_place_medal: .

awolfe76 (approved):

PR #584, opened 2017-06-01
Fix render & build bug

wpears:
- redux-oidc user state access needed to be protected
 - node-sass upgrade and travis node version bump


package.json
```diff
@@ -52,7 +52,7 @@
     "jest-cli": "18.1.0",
     "minifyify": "7.3.4",
     "mkdirp": "0.5.1",
-    "node-sass": "3.13.0",
+    "node-sass": "^4.5.3",
```
awolfe76: Do we want to use the `^` here?

wpears: In practice, it wouldn't matter as installs would proceed from the lockfile. I don't really like 4.5.3 of node-sass cause it doesn't support node v8.0.0 on linux, bizarrely and kinda kept the carat around as a reminder to upgrade.

However, updated for consistency.

awolfe76 (approved):

PR #585, opened 2017-06-01
Render simple message for verifier when no edits exist

### LINKED ISSUE #577: Don't show a verifier if no quality / macro exist
Or rather, have the verifier use the lack of edits to show a custom message

wpears:
Closes #577

Depends on #584


awolfe76:
I like this idea. With this update we see:

no-edits

What if we just use the "success" message above to get this point across? This would mean that the verifier would not be rendered and the "success" message would be updated with the "No verification required ..." content you have here.

Thoughts?

lesgou:
:+1: I like either one. Which would be more clear to filers that they would still have to select the button to proceed?

designlanguage:
Per Leslie's question, what if we flip the emphasis?

## Continue to the next section
#### No verification is required because no quality edits were found.

wpears:
Yeah I don't think something necessarily needs to be rendered (since nothing else is rendered for synval). Rendering the **No verification required** or **Continue to the next section** kinda assumes that the average filer will expect something to be rendered and will be perturbed/confused by its absence... I don't think this is the case.

If it is, we should probably render a **This section is complete, please continue** for *every* page (upload, synval, etc)

wpears:
Trying out the no markup route for now

awolfe76:
I like this.

no-quality

My only *suggestion* would be that we update that message to be something like:
"Your data did not trigger any quality edits, no verification is required."

Just to get mention verification since quality and macro require it ... except in this case.

awolfe76 (approved):
Nice!

PR #587, opened 2017-06-01
Show filing deadline

### LINKED ISSUE #543: Show the filing deadline in the header
(or somewhere prominent)

awolfe76:
Closes #543 

This is rendering as a banner, slightly different than the original proposal. It also display on if its between 1/1 and 3/1 (the filing period) and is just "informational" unless its within a week of the final day, when its a "warning".

# Informational

deadline-filing

# Warning

deadline-one-week


src/js/components/BannerDeadline.jsx
```diff
@@ -0,0 +1,44 @@
...
+    return true
+  }
+  return false
+}
+
+const BannerDeadline = (props) => {
+  const today = moment().format('YYYY-MM-DD')
+
+  // render if within the filing period (1/1/<filingPeriod> to 3/1/<filingPeriod>)
+  if(moment(today).isBetween(`${props.filingPeriod}-${dates.FILING_START}`, `${props.filingPeriod}-${dates.FILING_DEADLINE}`, 'day', '[]')) {
```
wpears: Won't this always be false because `today` will be the year *after* the filing period? The check should be on filingPeriod + 1

wpears (changes requested):
The one comment above needs to be addressed, also probably could use a test with `moment` mocked such that you can check the content of the entire rendered module

src/js/components/BannerDeadline.jsx
wpears: eg, you'll be running the app on 2/23/18, but the filingPeriod will be 2017

awolfe76: Good catch. Fixed in 7be8b08.

PR #593, opened 2017-06-07
Remove imgmin task from build

wpears:
imgmin seems a bit flaky across environments/versions, so I created an `img-dev` task to minify any new images we add and reduced the `img` task to just copying the pre-minified images over to `/dist`.

This will address certain cross-platform bugs we've encountered.


hkeeler (approved):
👍

Issue #569, opened 2017-05-26
Bring back the steps to the edits navigation


awolfe76:
We had these previously but they were removed (by me 😉 ) until we could get a better idea of how it should work.

## Early example (but we don't have to keep this style)

![edits-nav](https://cloud.githubusercontent.com/assets/2932572/26495356/483db080-41f1-11e7-92d3-f74fbffdff52.png)

designlanguage:
So I tried out a few things.

Initially, I was hoping to have the steps double as headlines for the sections, which would look something like this:

![screen shot 2017-06-07 at 1 07 13 pm](https://user-images.githubusercontent.com/10562544/26892545-af560b3e-4b86-11e7-81b1-e60494e967f4.png)

Speaking with @lesgou, if there are many types of edits (even with pagination), the remaining steps would leave the page, and that's not ideal.

The other two options have already been explored, but there are a couple details:

* I'd like to keep color on active/actionable areas to deemphasize completed or future items. In the steps, that would mean completed ones could be black, and future ones could be gray.
* The line behind the steps don't need to extend before the first step and after the last step. They can just connect the steps. They also don't need to change color; the focus can remain on the circles.

It could look like this:

![screen shot 2017-06-07 at 12 46 27 pm](https://user-images.githubusercontent.com/10562544/26894868-a9f4cb24-4b8d-11e7-9c9c-0bd60cc63a5e.png)

So the two options would be a persistent vertical navigation, or a horizontal one. I'd guess that vertical space is at more of a premium (especially if there are lots of edits), but the IRS might be constrained horizontally. 

| A | B |
| --- | --- |
| ![screen shot 2017-06-07 at 1 07 01 pm](https://user-images.githubusercontent.com/10562544/26895341-4ca8c414-4b8f-11e7-87ec-a6a0de8ec334.png) | ![screen shot 2017-06-07 at 1 07 25 pm](https://user-images.githubusercontent.com/10562544/26895340-4ca5a3ec-4b8f-11e7-91c2-798c929186f6.png) | 
| This matches the steps on the institutions page that has been proposed. There's a bunch of unused space under the steps though. | There's a bit less wasted space, but if vertical space is at a premium, this might not be ideal. |

I'm interested to hear preferences/feedback on this.

lesgou:
Would "B" be persistent as well? If I am going through the edits and am at the bottom of the page, would I still be able to see the steps?

keileyo:
[...]

PR #597, opened 2017-06-09
New validation progress

wpears:
Now with:
 - fancy (??) animations
 - progress bar
 - more tests


src/js/components/ValidationProgress.jsx
```diff
@@ -1,69 +1,61 @@
...
-  let textClass = ''
-
-  if(code < STATUS.VALIDATING) textClass = 'text-gray-light'
-  if(code === STATUS.PARSED_WITH_ERRORS) return <li className={textClass}>{getIndicator(code, 'validate')} Edit verification process will not run due to parsing errors</li>
-  if(code <= STATUS.VALIDATING) return <li className={textClass}>{getIndicator(code, 'validate')} Verifying edits ...</li>
-  return <li>{getIndicator(code, 'validate')} Edit verification complete</li>
+  return (
+    <div className="progressText">
+      <span>{text}</span>
+      {props.code === STATUS.VALIDATED_WITH_ERRORS ?
```
awolfe76: Can this just be another `if` on `props.code`? This messaging feels off to me:
edits

Maybe we could also call out "Edit validation complete, no edits found."

I think we should be consistent with the punctuation too and use periods everywhere.

src/sass/components/ValidationProgress.scss
```diff
@@ -1,33 +1,92 @@
...
+    background-color: $color-gray-lightest;
+    width: 100%;
+    height: 0.2em;
+    &::before {
+      content: "0";
+      font-size: 0.75em;
+      margin-top: -0.5em;
+      margin-left: 1.5em;
+      position: absolute;
+      left: 0;
```
awolfe76: I think if you remove this and add `bottom: 5px` and `font-weight: 700` it looks a bit cleaner.

src/sass/components/ValidationProgress.scss
```diff
@@ -1,33 +1,92 @@
...
+    }
+
+    &::after {
+      content: "100%";
+      font-size: 0.75em;
+      margin-top: -0.5em;
+      margin-right: -0.25em;
+      position: absolute;
+      right: 0;
```
awolfe76: Keeping this and adding `bottom: 5px` and `font-weight: 700` to match the `:before`.

src/sass/components/ValidationProgress.scss
```diff
@@ -1,33 +1,92 @@
...
-    background-color: $color-gray-light;
-    border-radius: 50%;
-    display: inline-block;
-    height: 10px;
-    margin-right: .25em;
-    width: 10px;
+  .progressFill {
+    background-color: $color-primary;
+    line-height: 1em;
+    height: 1em;
```
awolfe76, line 47: We should have the fill match the height of the progress total, so `.2em`.

Something like:
progress-bar

src/sass/components/ValidationProgress.scss
```diff
@@ -1,33 +1,92 @@
...
   }
-  .usa-unstyled-list>li {
-    margin-bottom: .5em;
+
+  .progressTotal {
+    background-color: $color-gray-lightest;
+    width: 100%;
+    height: 0.2em;
+    &::before {
+      content: "0";
```
awolfe76: Add the `%` here too.

awolfe76 (changes requested):
A few comments ... plus it would probably be a good idea to remove the progress from the header at this point too. Once we have the idea of steps back it's probably not needed anymore. (See examples in https://github.com/cfpb/hmda-platform-ui/issues/569#issuecomment-306894937)

src/sass/components/ValidationProgress.scss
wpears: I specifically removed it cause I thought it looked gross :racing_car: 

I can add it back in though

awolfe76: 😃  I think to be consistent with what we're showing it makes sense. I have a few thoughts in my head for some improvements too but need to play some.

src/sass/components/ValidationProgress.scss
wpears: I like the large progressFill, kinda lends more visual weight to what's been completed thus far.

src/sass/components/ValidationProgress.scss
wpears: I agree re: boldness, but I prefer the "either-side" positioning as opposed to below or above, particularly with the wider progressFill

src/sass/components/ValidationProgress.scss
awolfe76: I would make the gray background bar the same height then.

src/sass/components/ValidationProgress.scss
awolfe76: Not a really strong feeling here but something feels off with it sticking out past the div edges and doing that unevenly. Some symmetry gets lost.

src/js/components/ValidationProgress.jsx
wpears: Could be just another `if`, but the structure of what `text` get assigned to would need to change (eg. you'd have to repeat `span` everywhere)

Also, this is just to signal that validation has completed, not that no edits are found.

There was a note earlier that the red button made it unclear whether a file needed to be re-uploaded or if the user should move on to review elsewhere. The check here says "alright move on", but the strong text informs you everything isn't copacetic.

Also re: periods, I think I'd probably prefer to remove the one here than add them in (since that's more grammatically correct). I was trying to add some gravity to the "review required" bit with the period, but I think the period looks a little busy

awolfe76: I don't think you'd have to repeat the span if you did something like:
`if(props.code > STATUS.VALIDATED_WITH_ERRORS) text = <strong>Edits found, review required.</strong>`

But if it's not to signal that edits were found maybe we could just remove that bit. It looks strange to mean to have 2 sentence fragments next to each other and 1 bolded.

Maybe we could just modify the `NavButton` to get that messaging across on the upload page?

src/sass/components/ValidationProgress.scss
wpears: :wilted_flower: fiiiiiiiiiiiiiiiiiiiiiiiiiine :wilted_flower: 

😉

src/js/components/ValidationProgress.jsx
wpears: Moved the `strong` text to a new line.

Try that on for size 👍

awolfe76 (approved):

PR #602, opened 2017-06-14
Steps

### LINKED ISSUE #527: Progress indicator
After moving to the sticky header, we've lost most of the progress indicating-goodness from before (currently we only have a 1px line, which is pretty unclear if you don't know what you're looking for since it's also right on the dividing line between the header and the content below).

We should do some work to reinvigorate the progress indicator, which may include:
 - broadening the width of the indicator to 3-4px
 - separating the bar into segments to align with navbar items
 - pinning the width of the indicator to the width of the nav bar (so it doesn't extend too far past the first and last item)
 - reintroducing overlaid numbers to each segment to further emphasis progress
 - others

Thoughts @awolfe76 @designlanguage

### LINKED ISSUE #569: Bring back the steps to the edits navigation

awolfe76:
Closes #569 
Closes #527

# Screenshots

## Upload

nav-upload

## Uploaded

nav-uploaded

## Syntactical and Validity

nav-syn-valid

## Quality

nav-qual

## Macro

nav-macro

## Quality, with all edits verified

nav-qual-all-verified

## Confirmation

nav-confirm

## Confirmed

nav-confirmed

## Upload, signed

nav-all-complete

### DIFF
--- a/src/js/components/EditsNav.jsx
+++ b/src/js/components/EditsNav.jsx
@@ -5,8 +5,8 @@ import submissionProgressHOC from '../containers/submissionProgressHOC.jsx'
 import {
   PARSED_WITH_ERRORS,
   VALIDATING,
-  VALIDATED,
-  SIGNED
+  VALIDATED_WITH_ERRORS,
+  VALIDATED
 } from '../constants/statusCodes.js'
 
 const RefileWarning = submissionProgressHOC(RefileWarningComponent)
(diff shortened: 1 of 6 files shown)


src/js/components/EditsNav.jsx
```diff
@@ -153,17 +173,15 @@ export default class EditsNav extends Component {
...
-          </div>
+          <ul className="usa-nav-primary">
+            {
+              navNames.map((pageObj, i) => {
+                return renderLinkOrText(this.props, pageObj, i)
+              })
+            }
+          </ul>
+          <hr className="nav-bg" />
+          <hr />
```
wpears, line 184: I'm lukewarm on this `hr`, looks a bit busy to me.

wpears (changes requested):
Two issues:
 - text getting centrally justified in `RefileWarning`
 - Much too much whitespace between the nav and the start of the edits pages (which is also inconsistent between synval/quality/macro).

screen shot 2017-06-15 at 10 23 13 am

screen shot 2017-06-15 at 10 25 30 am

src/js/components/EditsNav.jsx
awolfe76: The only(?) benefit to the last `<hr />` is for when that goes sticky. It provides a break between the content and the nav in that case.

awolfe76:
@wpears - should be ok, now. Good catches!

PR #611, opened 2017-06-16
fix warning spacing and nav padding

### LINKED ISSUE #608: Fix spacing
There is some spacing between the edits nav and the warnings.

spacing

wpears:
Closes #608

### DIFF
--- a/src/sass/components/RefileWarning.scss
+++ b/src/sass/components/RefileWarning.scss
@@ -9,6 +9,7 @@
 
 .RefileWarning {
   animation: 0.3s ease-out 0s 1 animateWarningHeader;
+  border-top: 1px solid $color-gray-light;
   margin: 0;
   text-align: left;
 
--- a/src/sass/components/RefileWarning.scss
+++ b/src/sass/components/RefileWarning.scss
@@ -9,7 +9,7 @@
 
 .RefileWarning {
   animation: 0.3s ease-out 0s 1 animateWarningHeader;
-  border-top: 1px solid $color-gray-light;
+  border-bottom: 1px solid $color-gray-lighter;
   margin: 0;
   text-align: left;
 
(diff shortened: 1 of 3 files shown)


src/sass/components/RefileWarning.scss
```diff
@@ -9,7 +9,8 @@
 
 .RefileWarning {
   animation: 0.3s ease-out 0s 1 animateWarningHeader;
-  margin: 1px 0 0 0;
+  margin: -1px 0 0 0;
```
awolfe76: Instead of the negative margin here we can set the `hr` height to 0 in `EditsNav.scss`. (I think we can actually set the height to 0 in `hmda.scss`.)

src/sass/components/RefileWarning.scss
```diff
@@ -9,7 +9,8 @@
...
+  box-shadow: 0px 1px 3px -2px;
```
awolfe76: I don't think this is necessary. The background color here provides a border between the content and the warnings.

wpears: True, but I think it's a better practice to not rely on color alone and the 1px isn't much price to pay.

awolfe76: For the 1px a border would be a better solution.

awolfe76:
@wpears - I _think_ 7fc087a accomplishes the same thing without the extra `div`. Give it a look.

wpears:
That removes the divider between the EditsNav and the warning, which, imo, looks a little less polished that including the border.

awolfe76:
1fbfff5 should fix that.

wpears:
That method causes a flash on the warning's border when switching from a page with no warning to a page with a warning (upload something with just quality errors and switch from the upload page to the quality page)

awolfe76:
That should do it. (Still not sure that bottom border on the warning is worth it but its there.)

wpears:
:rocket: *c o m p r o m i s e* :rocket:

PR #614, opened 2017-06-21
Redirect to signin page when silent signin fails

### LINKED ISSUE #605: Paused progress on upload
After going to `/upload` I happened to let it sit for a while. When I came back and uploaded a file the progress started and then paused. It turned out I was logged out but there was no clear indication of this ... a refresh seemed to get me back to the upload screen (no file selected) and I was able to upload a new file.

wpears:
Closes #605 

Silent sign on is now leaned on a bit more heavily and when tokens cannot be renewed silently, users are redirected to the signin page.

### DIFF
--- a/src/js/actions/hasHttpError.js
+++ b/src/js/actions/hasHttpError.js
@@ -1,5 +1,5 @@
-export default function hasHttpError(json) {
-  return !json || json.httpStatus > 399 ?
-    true :
-    false
+export default function checkHttpError(json) {
+  if(!json) return Promise.resolve(true)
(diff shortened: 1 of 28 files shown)


src/js/actions/hasHttpError.js
```diff
@@ -1,5 +1,5 @@
-export default function hasHttpError(json) {
-  return !json || json.httpStatus > 399 ?
-    true :
-    false
+export default function checkHttpError(json) {
```
awolfe76: Do you think we should update the filename here too? And maybe the imports then too, just to be clear?

PR #615, opened 2017-06-22
Current filing CSV download

### LINKED ISSUE #530: Make edit report CSV download always available
We should have the edit report CSV download available from all submission pages. Possibly fixed along with the edit navigation.

awolfe76:
Closes #530 

This doesn't make the download always available but it does make it available on the institution listing, which I think is enough at the moment. Some usability testing can tell us if it's important to provide in on the edit pages.

I updated the formatting of the messaging to make a little more sense, _I think_. 😄 

download-csv

### DIFF
--- a/__tests__/components/Institutions.js
+++ b/__tests__/components/Institutions.js
@@ -14,6 +14,7 @@ import React from 'react'
 import ReactDOM from 'react-dom'
 import TestUtils from 'react-addons-test-utils'
 
+
 const fs = require('fs')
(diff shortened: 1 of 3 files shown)


lesgou (commented):
I like this version of the messaging, and agree that we don't need the csv for beta (but may need it for next phase). Will wait for @designlanguage and @keileyo to comment to merge.

awolfe76:
And don't forget, @wpears needs to review the code. 😉

wpears (approved):
Approved, with #618 noted

PR #620, opened 2017-06-23
fix for IRS continuing to poll even after response

awolfe76:
IRS was continuing to poll even after a valid response.


wpears:
Nice catch ^_^

PR #622, opened 2017-06-23
Pagination transitions

wpears:
Animates both parse errors and edits when paginating

:eyes:


src/js/components/Pagination.jsx
```diff
@@ -19,6 +21,7 @@ const Pagination = (props) => {
         className={ lastPage ? 'usa-button-disabled' : '' }
         onClick={ e => { if(!lastPage) props.getNextPage(page) }}
       >Next</button>
+      {props.isFetching ? <LoadingIcon/> : null}
```
awolfe76, line 24: Would this be better to show in the results, eg parsing errors or edits? Or does it cause more jumping around the page?

awolfe76:
I see you started with a fade approach to this, I think I'd like to see that one too. It seems simpler for the UI (and for the code). Google takes this approach on their search results and the first example [here](https://css-tricks.com/animating-the-content-property/) (even though we wouldn't do it with those properties and we'd likely want to be faster).

And did you check out https://facebook.github.io/react/docs/animation.html?

wpears:
Transition groups can't be easily used in connected components (there's kinda a messy patch that I played around with but decided against).

https://github.com/chenglou/react-motion is an option but seems like the wrong fit to me, really we just want pretty basic animations triggered at the correct times.

A version with sliding removed and replaced with fading has been introduced at #628

PR #621, opened 2017-06-23
Irs csv

awolfe76:
Closes #517

PR #625, opened 2017-06-26
Error handling bug

### LINKED ISSUE #623: Ensure error messaging
It has been observed that if the backend crashes validating and the institutions poll returns a 504, a error warning won't show. If an error is (ever) encountered, there should always be a friendly warning for the user.

wpears:
Fixes and generalizes error handling

Addresses a bug found in https://github.com/cfpb/hmda-platform-ui/pull/621 where a non-200 response to the CSV endpoint would result in a CSV of the error being downloaded.

Closes #623


awolfe76 (approved):
👍

PR #628, opened 2017-06-27
Fade transitions

wpears:
Alternate approach to #622 

Note that there is no placeholder when waiting for data, so there is a large swath of whitespace and a spinner. This *could* be changed such that the current data stays until the next is available, with a fade out/in upon receiving it, but with a hit to perceived responsiveness


awolfe76:
I like this one. It feels a lot less distracting than the sliding.

> but with a hit to perceived responsiveness

Wouldn't that potentially apply in any case? It's always possible to have the API response be a bit slow, we just need to handle that for any type of animation where it depends on the API.

In this case, with the fade, when they click to page maybe

- first, fade out to 50% the current results
- when the response is received, fade out the rest of the way
- fade in the new results

Doing the first fade at least gives an idea something happened. I'm sure there are ways to improve it more, but seems like a good first step.

wpears:
Some general fixups here, `td`s now fade (which I thought could lead to jank but looks pretty good). This allows the table rule to be it's own placeholder (it fades to `0.5` to not be overwhelming).

:eyes:

src/sass/components/Pagination.scss
```diff
@@ -26,7 +26,16 @@
...
   transition: opacity 0.3s;
 
+  td {
+    opacity: 1;
+    transition: opacity 0.3s;
+  }
   &.fadeOut {
-    opacity: 0;
+    opacity: 0.5;
+    box-shadow: 0 1px 1px -2px;
```
awolfe76: This doesn't seem to do anything. Or am I missing it?

awolfe76 (commented):
Just the one small thing about the `box-shadow` but this is good to go. I like it!

src/sass/components/Pagination.scss
wpears: Without the box-shadow, the bottom border thins by ~1px on fade in a somewhat distracting manner. By adding it in, it APPEARS to do nothing, which is exactly what we want ^^;

awolfe76: Bottom border of the table? I tried with and without and didn't notice anything different.
If you think it needs to be there maybe just a quick comment to tell why.

PR #631, opened 2017-06-28
improve semantics

### LINKED ISSUE #599: Semantic tags
Following the example in #598, add semantic tags throughout the application.

awolfe76:
Closes #599 

Open to other suggestions/recommendations here.


src/js/containers/Submission.jsx
```diff
@@ -107,11 +107,11 @@ class SubmissionContainer extends Component {
 
       <main id="main-content" className="usa-grid SubmissionContainer">
         {this.props.error ? <ErrorWarning error={this.props.error}/> : null }
-        <div className="usa-width-one-whole">
-          {toRender.map((component, i) => {
-            return <div key={i}>{component}</div>
-          })}
-        </div>
+        {toRender.map((component, i) => {
+          console.log(i)
```
wpears: Stray logs

wpears (changes requested):
I'm somewhat mixed on the use of `article` throughout this PR. I think it *might* make sense for the edit tables, but I don't think it's justified anywhere else. [According to the w3c](https://www.w3.org/wiki/HTML_structural_elements#.3Carticle.3E), an `article` should be able to stand on it's own and would make sense to syndicate, which I don't think applies to *anything* in our app (aside from edit tables which someone could argue represent a complete piece of information about one edit, but imo don't make sense without user context, and other ancillary information on the page.

I don't think we'll lose too much by reverting these to `div`s, especially since most of them are wrapped inside `section`s

awolfe76:
Makes sense to me. I was a bit "mixed" on it too. The difference in `<article>` and `<section>` isn't the clearest, except for the stand-alone argument. I'll dig through again.

awolfe76:
Removed logging and changed back _most_ of the `<article>` tags. I left a few behind ...

- home page faqs
- banner deadline (filing deadline)
- edits, like you mentioned

I think those few things fit the spec.

PR #632, opened 2017-06-29
Trim auth flows

### LINKED ISSUE #624: Incorrect routing on token retrieval
Occasionally (need solid repro), after refresh a submission will be sent to the upload page, no matter their target page.

Seems to be a problem with `restorePage`

wpears:
Closes #624 

Removes odd edge case behavior, unnecessary re-renders, and subtle visual bugs.

### DIFF
--- a/src/sass/components/Header.scss
+++ b/src/sass/components/Header.scss
@@ -1,0 +1,12 @@
+.Header {
+  nav {
+    margin: 0;
+    padding: 0;
+  }
+  .usa-nav-primary {
+    .usa-button {
+      width: 5.5em;
+      padding: 1rem 0;
+    }
+  }
+}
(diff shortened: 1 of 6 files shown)


src/sass/components/Header.scss
```diff
@@ -0,0 +1,12 @@
+.Header {
```
awolfe76, line 1: Turning these off renders nearly (if not completely) identical ... or am I missing something?

wpears: When refreshing the page the blue `Login` button initially is slightly wider than it's automatic width then shrinks. It's only on page load and lasts for <100ms but happens every time.

awolfe76: Feels like an odd fix for such a small thing ... and I'm not seeing that happen at all.

wpears: the margin and padding aren't actually required for the fix, but they help collapse the wrapper to the content, which I think it appropriate here.

The other change is just setting a defined width so the `auto` width doesn't recalculate during load.

PR #640, opened 2017-06-30
Hide rather than disable nav button

### LINKED ISSUE #629: Nav button on upload page
We are currently showing the nav button when the upload page is initially loaded:

upload

The _potential_ problem I see with this is when parser errors occur. In this case we see:

parse-errors

When we didn't have the idea of steps in the edits navigation showing this button initially was more important. Now, in the case of parsing errors, it will never be enabled.

Would it be better to not show the button on the upload page initially and only render it when a file successfully parses (filer can move beyond the upload page)?

wpears:
Closes #629

### DIFF
--- a/__tests__/components/NavButton.js
+++ b/__tests__/components/NavButton.js
@@ -13,8 +13,8 @@ const baseProps = {
   macroVerified: false
 }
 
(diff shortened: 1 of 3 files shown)


awolfe76 (approved):
👍

Issue #637, opened 2017-06-29
Pagination: navigate to each page


wpears:
Certain filers may have dozens or even hundreds of pages. They should be able to travel at will to any page.

awolfe76:
Closed by #647

PR #647, opened 2017-07-05
Arbitrary pagination

wpears:
Enables ability to page to any page in a given edit's range.
 - Doesn't submit on non-ints
 - Brackets values to ints within range (and updates input accordingly)


awolfe76:
Closes #637

PR #656, opened 2017-07-13
Break out the alerts into a component

### LINKED ISSUE #633: Alerts could be a component
We use [alerts](https://standards.usa.gov/components/alerts/) throughout the application, we should think about extracting this into its own component.

awolfe76:
Closes #633 

Left out the [`UploadForm`](https://github.com/cfpb/hmda-platform-ui/blob/10a3ea35f52169560cada2fdcf0a913376a813a2/src/js/components/UploadForm.jsx) because of the way it [renders the possible list of errors](https://github.com/cfpb/hmda-platform-ui/blob/10a3ea35f52169560cada2fdcf0a913376a813a2/src/js/components/UploadForm.jsx#L23). Any thoughts there?

_Some minor formatting updates._

### DIFF
--- a/src/js/components/Alert.jsx
+++ b/src/js/components/Alert.jsx
@@ -1,7 +1,7 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 
-const Alert = ({ type = 'info', heading, text }) => {
+const Alert = ({ type = 'info', heading, text, htmlElement }) => {
   if (!text) return null
 
   return (
@@ -9,15 +9,17 @@ const Alert = ({ type = 'info', heading, text }) => {
       <div className="usa-alert-body">
         {heading ? <h3 className="usa-alert-heading">{heading}</h3> : null}
         <p className="usa-alert-text">{text}</p>
+        {htmlElement ? htmlElement : null} {/* allow for additional html */}
       </div>
     </div>
   )
 }
 
 Alert.propTypes = {
-  type: PropTypes.string,
+  type: PropTypes.oneOf(['info', 'success', 'warning', 'error']),
   heading: PropTypes.string,
-  text: PropTypes.string.isRequired
+  text: PropTypes.string.isRequired,
+  htmlElement: PropTypes.node
 }
 
 export default Alert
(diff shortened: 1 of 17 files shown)


src/js/components/Alert.jsx
```diff
@@ -0,0 +1,25 @@
...
+
+const Alert = ({ type = 'info', heading, text, htmlElement }) => {
+  if (!text) return null
+
+  return (
+    <div className={`usa-alert usa-alert-${type}`}>
+      <div className="usa-alert-body">
+        {heading ? <h3 className="usa-alert-heading">{heading}</h3> : null}
+        <p className="usa-alert-text">{text}</p>
+        {htmlElement ? htmlElement : null} {/* allow for additional html */}
```
wpears: Instead of passing in the custom `htmlElement` prop, you can just use `props.children` and wrap children in standard html.

wpears: ie
```
<Alert some={props}>
  <p>Like this</p>
</Alert>
```

wpears (changes requested):

src/js/components/Alert.jsx
```diff
@@ -1,15 +1,15 @@
...
+const Alert = ({ type = 'info', heading, text, children }) => {
   if (!text) return null
 
   return (
     <div className={`usa-alert usa-alert-${type}`}>
       <div className="usa-alert-body">
         {heading ? <h3 className="usa-alert-heading">{heading}</h3> : null}
         <p className="usa-alert-text">{text}</p>
-        {htmlElement ? htmlElement : null} {/* allow for additional html */}
+        {children ? children : null}
```
wpears: afaik `props.children` will always either be a valid element or `null`, so you don't need this check here

wpears:
@awolfe76 so I missed it the first time (somehow), but the method of passing an array of nodes as a prop would require a `key` on each one, which kinda bungles it.

I reimplemented this approach by just cloning the node that Alert wraps with the proper className. This also makes the structure of children more obvious, imo, wrt to how it interacts with the text of the alert (since you're just passing children as children and not as props that get rendered as children).

I did still allow the heading to be specified as a prop, since this allows it to be kept a bit more separate.

wpears (approved):

awolfe76:
👍

PR #667, opened 2017-07-21
Update redirect uri to always https version

wpears:
Companion to https://github.com/cfpb/hmda-platform/pull/1049 and https://github.com/cfpb/hmda-platform-auth/pull/126

Since the app is now always https in compose, we need to update the `redirect_uri` (as provided by the `APP_URL` envvar


wpears:
ping @hkeeler

hkeeler (approved):
👍

Issue #658, opened 2017-07-13
Messaging and app state for resubmitting signed filings


wpears:
Currently, filings have three states
  - not-started
  - in-progress
  - completed

They start in the first state, move to the second state when a submission is created, and the last state when a filing is signed.

However! If an FI wants to *resubmit after signing* the filing status *doesn't change back to in-progress*. While this may seem like a backend bug, it's actually expected behavior, because, legally (IANAL), they've still submitted a filing and even if they quit out of their resubmission without signing it, their first submission still counts... so it **perhaps** makes sense to keep the filing in the completed state even if they start a resubmission.

This would mean we would need some strong messaging here about what is going on in the app, eg, "hey, looks like you're resubmitted after signing a filing. Your current signed filing will be counted for YYYY filing year unless/until you complete and sign this resubmission."

Implementation-wise, this will just be checking for filing-status of of 3/completed, with any submission status other than 10/signed.

**Alternatively** we could add a new state to the filing (eg `resubmission-in-progress`) or a flag to the submission state (eg `isResubmission`) and then key off that in front-end code to message things, rather than relying on the combination of completed and a non-signed submission status. I'd probably prefer this latter approach as it seems to explicitly capture more of the actual submission/filing status rather than relying on implicit knowledge of the current odd combination of status codes, but I have no problem soldiering on ahead with the current setup.

As to messaging, input would be much appreciated @designlanguage @keileyo @lesgou @awolfe76 @schbetsy 
Note that this will have effects on what is shown on the filing page and the banner warnings for when near the filing date (and most likely other things as well)

Thanks~

lesgou:
To make things more complicated, a "resubmission" doesn't actually count as a resubmission until it has been signed and submitted (in the business world). On the code front, I think it makes sense to call it resubmission in progress, but on the messaging front, how about:

"You are about to upload a new file to an already submitted filing. This new HMDA file will not be submitted until you clear the edits, sign and submit the data."

PR #674, opened 2017-07-28
Key view button off submission status rather than filing status

wpears:
Handles much of #658

### DIFF
--- a/src/js/components/Institutions.jsx
+++ b/src/js/components/Institutions.jsx
@@ -68,14 +68,12 @@ export const renderStatus = (
 }
 
 export const renderViewButton = (code, institutionId, period) => {
-  let buttonText
+  let buttonText = 'View current filing'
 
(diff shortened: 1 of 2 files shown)


src/js/components/Institutions.jsx
```diff
@@ -70,25 +70,12 @@ export const renderStatus = (
...
-    case 4:
-      buttonText = 'Begin filing'
-      break
-    default:
-      buttonText = 'Begin filing'
+  if(code === STATUS.CREATED){
+    buttonText = 'Begin filing'
+  }else if(code === STATUS.SIGNED){
+    buttonText = 'View completed filing'
+  }else{
```
awolfe76: Can we remove the last `else` by setting `let buttonText = 'View current filing'` above?

awolfe76 (approved):

PR #687, opened 2017-08-01
Defer to store for data where appropriate

wpears:
Rather than going to network, first check whether the state in the store is fresh/not the default.

Also, this reworks the institutions component a bit to use the actual submission status, rather than the proxied version available in the filings object (which doesn't get refreshed on status updates).


awolfe76 (approved):
🎆

PR #688, opened 2017-08-02
Banner USA before beta

wpears:
I think it makes more visual sense


wpears:
This is actually in line with beta cf.gov, so **gross as it is** gonna keep it with beta banner first

PR #689, opened 2017-08-02
Fix bug when no submissions exist

### LINKED ISSUE #660: Institution with no submissions
The UI should have some messaging for when an institution first enters the application. In this case, the institution will have the filing created, but no submissions. It currently looks like:

fi-no-submission

Other cases, after a submission has been triggered, look like (notice the status message):

fi-with-submission

@designlanguage @lesgou @keileyo

wpears:
(now closes #660)


awolfe76 (approved):
👍

PR #691, opened 2017-08-02
Sluggish check

### LINKED ISSUE #682: Checkboxes are waiting on results to check/uncheck
Makes the app feel sluggish. Instead, the check should be instant and a spinner should load as the backend thinks.

wpears:
Closes #682

Verifier check now activates instantly and shows a spinner when loading.


awolfe76 (approved):
👍 

I do think there are [some things](https://github.com/cfpb/hmda-platform-ui/issues/682#issuecomment-319407518) to think about here still but this works.

PR #704, opened 2017-08-07
Remove upload button

### LINKED ISSUE #695: Remove the inactive/gray "upload" button
Several people have tried to click on this first. We should remove it to focus attention on the drag/drop box. Blue upload button should appear after file is selected.

wpears:
Closes #695

Depends on #694

### DIFF
--- a/__tests__/components/UploadForm.js
+++ b/__tests__/components/UploadForm.js
@@ -12,7 +12,6 @@ import ReactDOM from 'react-dom'
 import TestUtils from 'react-addons-test-utils'
 
 describe('submitform', function(){
-  const handleSubmit = jest.fn()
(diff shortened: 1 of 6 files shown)


awolfe76 (changes requested):
Should we give some indication in the [text](https://github.com/cfpb/hmda-platform-ui/pull/704/files#diff-957168b6d48fd8a546ca3cf576065615R33) that the file will begin to upload immediately?

Instead of `To select a file to upload` maybe `To begin the file upload`?

Or, another sentence added to the existing content? So `To select a file to upload, drag it into this box or click here. The upload process will begin immediately after.`

awolfe76 (approved):

PR #707, opened 2017-08-07
Give logo a hover state and an underline

### LINKED ISSUE #701: Make "HMDA filing" link look more actionable
Tests revealed that some people didn't realize they could click on "HMDA filing" to return to the home/institutions page.

For an attempted quick-fix, can we underline it and give it a rollover color?

We've also referred to this as the "HMDA Platform". Shouldn't we use that wording instead (@keileyo @lesgou)?

wpears:
Closes #701 
(with https://github.com/cfpb/hmda-platform-auth/pull/141 to stay consistent)


awolfe76 (changes requested):
From #701 it looks like this should be updated to say `HMDA Platform`?

awolfe76:
Also see [this comment](https://github.com/cfpb/hmda-platform-ui/issues/701#issuecomment-321587805). Seems like that might be an even better solution.

wpears:
@awolfe76 Underline removed and homelink added

awolfe76 (approved):
It _might_ make more sense to have the home link to the left of the user/logout but lets give it a try and find out. 😄

PR #708, opened 2017-08-08
Jumping table pagination

### LINKED ISSUE #698: In pagination, location sometimes jumps
Clicking "next" on a paginated table sometimes jumps you to the bottom of the page. This behavior has been captured in a couple user tests, if it isn't easily replicable.

wpears:
- Little table redesign / cleanup (a little less padding and a lighter rule
- Paging to the last page no longer causes a jump due to some scrollTo hackery

Depends on #704 
Closes #698


src/sass/hmda.scss
```diff
@@ -32,6 +32,11 @@ hr {
     padding: 0;
 }
 
+table td,
+table th {
+  padding: 1rem;
```
awolfe76: What do you think about shrinking this even more? `.75` seems nice too ... and seems to make a bigger improvement to the IRS at least.

Also, in either case, [this](https://github.com/cfpb/hmda-platform-ui/blob/f6e3368d976fcc78a895c34ecd8b4877326a39a2/src/sass/components/IRSReport.scss#L20) can be removed.

And in the IRS file, we should add `tfoot td` [here](https://github.com/cfpb/hmda-platform-ui/blob/f6e3368d976fcc78a895c34ecd8b4877326a39a2/src/sass/components/IRSReport.scss#L18). I noticed the font size is off there.

src/sass/components/ParseErrors.scss
```diff
@@ -1,4 +1,9 @@
 .ParseErrors {
+  table td:first-of-type,
```
awolfe76, line 2: 👍

awolfe76 (changes requested):
Just a few small things here.

wpears:
@awolfe76 agreed re: padding, just didn't want to make quite so big a change all at once with some eyes on it. Updated.

awolfe76 (approved):

PR #709, opened 2017-08-09
Make edits nav check clickable

### LINKED ISSUE #703: Make edits nav checks clickable
To prevent Hans' sadness

wpears:
Closes #703


awolfe76 (changes requested):
I'm noticing some alignment issue for the numbers in different states.

@wpears - let me know if you'd like me to mess around with this a bit.

wpears:
Sure, go nuts

awolfe76:
This didn't quite get rid of the misalignment but did help. It also removed some of the markup.
Still has the steps as links though. 😄 

I'm going to merge so we can carry on with the other PRs. If we need to we can revisit later.

PR #714, opened 2017-08-10
Safari bug

### LINKED ISSUE #713: Bug in Safari
When trying to run the app using Safari you'll see ...

```
Error: Error calculating filing deadline status
withinAWeekOfDeadline — app.min.js:44020
BannerDeadline — app.min.js:39024
```

wpears:
Closes #713

### DIFF
--- a/src/sass/components/UploadForm.scss
+++ b/src/sass/components/UploadForm.scss
@@ -58,10 +58,9 @@
     }
 
     button {
-      align-items: center;
-      display: flex;
+      display: block;
       height: 100%;
-      justify-content: center;
+      text-align: center;
       line-height: normal;
       margin: 0;
       margin-right: -50%;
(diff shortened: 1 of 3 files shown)


src/sass/components/UploadForm.scss
```diff
@@ -58,10 +58,9 @@
     }
 
     button {
```
awolfe76, line 60: We have 2 `button` in this file, both within `.dropzone` ... see line 71.
Can we bring these together in this PR?

awolfe76 (changes requested):
Just the one thing and good to go.

src/sass/components/UploadForm.scss
wpears: good to go

awolfe76 (approved):

PR #717, opened 2017-08-16
Progress estimator

### LINKED ISSUE #642: Improve upload/validation progress bar and messaging
The progress bar could be improved, both visually and UX. 

Some things to think about (there could be others):
- positioning on the `%` on either side (are they even necessary in this case)
- positioning of the messaging for each stage
- the blinking blue dot
- should we show an additional % for the current progress, that moves along with the blue bar
  - could we show the messaging along with this
- we show all the steps in the submission process, should we do the same here

# In progress
progress-in-progress

# Complete
progress

### LINKED ISSUE #697: Remove or replace "Invalid file format" on parsing errors page.
When a file with parsing errors is uploaded, there's a message under the progress bar that says "invalid file format." That wording sounds makes it sound like it's not a text file. I recommend we either remove this wording (since formatting errors are noted below this and at the top of the page) or replace it with something more consistent, like "File contains formatting errors".

screen shot 2017-08-04 at 11 25 46 am

wpears:
Closes #697 
Closes #642

### DIFF
--- a/src/js/components/ValidationProgress.jsx
+++ b/src/js/components/ValidationProgress.jsx
@@ -67,14 +67,20 @@ export default class ValidationProgress extends Component {
     return <div className={className} style={{width: currWidth+'%'}}></div>
   }
 
-  getNextWidth() {
-    const currWidth = this.state.fillWidth
-
-    this.timeout = setTimeout(() => {
+  setNextWidth(currWidth) {
+    return () => {
       let nextWidth = currWidth + 1
       if(nextWidth > 100) nextWidth = 100
       this.setState({fillWidth: nextWidth})
-    }, SCALING_FACTOR*200 * Math.pow(2, 50/(100 - currWidth)))
+    }
+  }
+
+  getNextWidth() {
+    const currWidth = this.state.fillWidth
+    this.timeout = setTimeout(
+      this.setNextWidth(currWidth),
+      SCALING_FACTOR*200 * Math.pow(2, 50/(100 - currWidth))
+    )
   }
 
   componentWillUnmount() {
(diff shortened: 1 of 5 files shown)


src/js/components/ValidationProgress.jsx
```diff
@@ -1,61 +1,107 @@
...
-  return <div className={className} style={{width:fillWidth+'%'}}></div>
-}
+    return (
+      <section className="progressText">
+        <span>{text}</span>
+        {this.getIndicator()}
+        {code === STATUS.VALIDATED_WITH_ERRORS ?
+          <strong>Edits found, review required</strong> :
+          largeFile && code !== STATUS.PARSED_WITH_ERRORS ?
+            <strong>This may take several minutes</strong> :
```
awolfe76, line 41: I think after the upload is complete we should tell them that it's ok to leave here too.

awolfe76 (changes requested):
When I tried this with `edits-M_5508-lars` it stayed at ~10% during the whole process then jumped to 100% when complete. Should it be slowly progressing?

src/js/components/ValidationProgress.jsx
wpears: Is that something people are specifically asking for/confused about? Seems like it wouldn't really make the process easier for them and of course they can just open a new tab, etc.

wpears:
Bar should move more naturally now, got bit by a last minute change cleaning up the timeout.

src/js/components/ValidationProgress.jsx
awolfe76: @designlanguage seemed to think that it would be helpful to filers to understand that they didn't have to sit and wait and watch. Wouldn't mind hearing from him first. I'm ok either way.

awolfe76: Created #724 to catch this.

awolfe76 (approved):

PR #723, opened 2017-08-17
Irs bug

### LINKED ISSUE #650: API bug causes missing IRS table
There's a bug in the API where the IRS endpoint sometimes returns an empty result (see sample response below). When this happens, the UI currently doesn't show any IRS table or give any hint that there should be something in that section. Maybe we should inform the user that there's been an error?

Here's one way I've seen this happen:

1. Start the application with docker-compose
2. Create an account and log in
3. Upload a file (I used `edits-M_1124-lars.txt`, but I've seen it happen with others)
4. Verify Macro errors, proceed to Summary/Submission page
5. See that the IRS table isn't there
6. Check the "Network" tab of the developer console. See that the call to the irs endpoint had this response:

```json
{"count":0,"total":0,"msas":[],"summary":{"homeImprovement":0,"multiFamily":0,"lars":0,"FSA":0,"FHA":0,"amount":0,"oneToFourFamily":0,"refinance":0,"MFD":0,"conv":0,"homePurchase":0,"VA":0},"_links":{"self":"?page=1","prev":"?page=1","last":"?page=0","next":"?page=0","first":"?page=1","href":"/institutions/6999999/filings/2017/submissions/1/irs{rel}"}}
```

In my observation, it seems to happen on the first file submitted after starting the app. Needs confirmation.

awolfe76:
Closes #650 

Protects us a bit against https://github.com/cfpb/hmda-platform/issues/1031

Still allows filer to sign, signature doesn't require IRS, but gives them a message to contact us to get it.

### DIFF
--- a/src/js/components/IRSReport.jsx
+++ b/src/js/components/IRSReport.jsx
@@ -2,108 +2,156 @@ import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import Pagination from '../containers/Pagination.jsx'
 
-const renderTotals = (props) => {
-  if(props.renderTotals) return (
-    <tfoot>
-      <tr className="totals">
-        <td className="center" colSpan={2}><strong>Total</strong></td>
-        <td>{props.summary.lars.toLocaleString()}</td>
-        <td>{props.summary.amount.toLocaleString()}</td>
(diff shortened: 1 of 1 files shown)


wpears:
Cool beans

PR #736, opened 2017-08-23
Prev filename bug

### LINKED ISSUE #728: Filename exists on first filing
After I rebuilt the local setup, registered an account, logged in, and clicked begin filing I saw this 

filename.

Notice the filename, but I hadn't done anything yet. This was the first time through.

wpears:
Closes #728

### DIFF
--- a/src/js/containers/Submission.jsx
+++ b/src/js/containers/Submission.jsx
@@ -20,6 +20,7 @@ import RefileButton from '../containers/RefileButton.jsx'
 import ParseErrors from './ParseErrors.jsx'
 import LoadingIcon from '../components/LoadingIcon.jsx'
 import {
+  CREATED,
   UNINITIALIZED,
   FAILED,
   PARSED_WITH_ERRORS,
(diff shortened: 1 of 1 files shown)


awolfe76:
@wpears I had the same thought as [you did here](https://github.com/cfpb/hmda-platform-ui/issues/728#issuecomment-324464059) but I think it could be a little bigger than that.

If you have multiple filers from the same institution doing things (leaving, coming back, reuploading) it's possible that filename won't be correct since we're using `localStorage` here.

We may need to hit the `summary` endpoint to get the filename based on part of the `state`?

wpears:
File names aren't going server-side at all, so unless we change the API and make the filename part of the upload POST (a header like X-HMDA-Filename) that gets associated with a submission, this is the best we can do.

We'd still run into an issue if:
 - User A uploads
 - User B refiles
 - User A returns and sees "`name_of_first_file.txt` in progress" on the upload page, if they venture there.

This is pretty edge-casey, but would likely result in some confusion.

I could key the localStorage by both id AND submission, which would fix this issue. Then we'd just revert to `Your submission is in progress` and show no filename at all.

Thoughts?

awolfe76:
We aren't explicitly sending it back, as you suggest, but the [back-end does have it](https://github.com/cfpb/hmda-platform/blob/master/Documents/API.md#summary). We use that to render the summary information on the confirmation page. (And that page could cause even more confusion ... "I uploaded `x.txt` and the upload page says so, so why is the summary saying `y.txt`?")

So for this case, where 2 filers are working and overwriting each other, that endpoint could help.

wpears:
Oh duh, it's in the Content-Disposition of the upload. It would be pretty keen to associate this filename with the submission then, if we're already using it elsewhere.

awolfe76 (approved):

PR #741, opened 2017-08-24
Desc nav

### LINKED ISSUE #700: More descriptive nav bar
In order to limit the amount of vertical space taken up by the error message, some of that indication can be shown in the nav bar. The text and symbol can change on the steps depending on what has taken place.

Instead of the nav saying "quality edits" maybe is says "quality edits found" when you get to that step, and "quality edits validated" when they are checked.

screen shot 2017-08-04 at 11 58 58 am

awolfe76:
Closes #700 

## Screenshots

nav1

nav2

nav3

nav4

nav5

nav6

nav7

@designlanguage - the "uploaded with formatting errors", the last image, should it be red? The way it works right now is the current page is always blue, and the formatting errors render on the upload page so it would __always__ be blue.


wpears:
If we made "uploaded with formatting errors" red we'd also want to keep syn/val red upon visiting and probably quality and macro orange.

But yeah, seems like we're using the blue to indicate current page, so either we'd need to do that in a different way (underline, etc.) or just keep it how it is and assume that the current page color is more important to show than the "something is wrong" color

awolfe76:
Agreed, we should probably come up with a new way to show "current" and keep the colors. I think the colors make sense. I'll play around with some ideas and share.

awolfe76:
@wpears - 344f2b4 adds a bottom-border for current. I do think it could be better but this should get the point across for now.

PR #745, opened 2017-08-31
Refactor institutions

### LINKED ISSUE #618: Refactor Institutions component
Over time this has ballooned and the function signatures of the helper components have gotten gross.

This needs to be split into a few modules and common functionality should be factored out.

awolfe76:
Closes #618


awolfe76:
@wpears - this should be ready. I think we can do even more here (create a few more components, create containers for a few) but this should be a decent first step.

src/js/components/InstitutionsHeader.jsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+const InstitutionsHeader = props => {
+  return (
+    <header>
+      <h1>{props.filingPeriod} filing period</h1>
+      <p>The filing period is open. You may file HMDA data for your authorized institutions below.</p>
```
wpears: This clashes with https://github.com/cfpb/hmda-platform-ui/blob/9e9817fae3694b834046d0a2260abfd65987b431/src/js/components/Institutions.jsx#L11-L25

src/js/components/Institutions.jsx
```diff
@@ -1,297 +1,101 @@
...
-                          filingObj.submissions,
-                          this.props.onDownloadClick,
-                          institution.id,
-                          filing.period
-                        )}
-                      </section>
-                    </div>
-                  )
-                })}
+          <InstitutionsHeader filingPeriod={this.props.filingPeriod} />
```
wpears: The logic in the render here is still a bit opaque. Maybe some helper functions could replace the inline stuff
A `renderBody` fn could call an `isLoading` and a `resolveSubmission` function etc. to mask/separate this logic a bit

src/js/components/InstitutionNameAndId.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+const InstitutionNameAndId = ({ name, id }) => {
```
wpears, line 4: This component feels a bit unnecessary, like it could have just been a `renderNameAndId` function in the Institution component.

That said, I think stylistically you were going for a super simple institution component that just passes through data and doesn't handle render logic outside of organizing the proper components, so I'm okay with it. Also, the nameAndId component could grow/change and that would be masked in this approach.

wpears (changes requested):
Overall looks good, just the conflict in whether the filing period is open/closed and some ideas on how to make the core `Institutions.jsx` logic a bit less fraught.

src/js/components/InstitutionNameAndId.jsx
awolfe76: Exactly, I would like us to do more of this ... breaking down components into smaller, "dumber" components. It comes back to the ["Thinking in React"](https://facebook.github.io/react/docs/thinking-in-react.html) article. Over time we've moved away from that and IMO it would be good to start moving back towards it.

This is also why I think we will, eventually, need to think about how our [`/src` is structured](https://medium.com/styled-components/component-folder-pattern-ee42df37ec68). By breaking things out file names become _too important_, having them in a directory helps a lot.

Just as an example (might not be necessary in this case but could be done), notice the `Institution` prefix already on the new components. And if we break out `InstitutionPreviousSubmissions` anymore you end up with things like `InstitutionPreviousSubmissionsButton` and `InstitutionPreviousSubmissionsList`, etc.

src/js/components/InstitutionsHeader.jsx
awolfe76: Fixed in b384fc7

src/js/components/Institutions.jsx
awolfe76: c387102 does a bit here. I like that the actual renders all happen within `render()` this way, leaving the `.map()` there. We can always do more later.

src/js/containers/Institutions.jsx
```diff
@@ -24,15 +22,9 @@ export class InstitutionContainer extends Component {
 export function mapStateToProps(state) {
   const { institutions } = state.app.institutions
 
-  const {
-    filings,
-    submission,
-    error,
-    filingPeriod
-  } = state.app
+  const { filings, error, filingPeriod } = state.app
```
awolfe76: @wpears - this bit is new. While doing some testing it seemed like `submission` never existed at this point. Plus we [can do this](https://github.com/cfpb/hmda-platform-ui/pull/745/files#diff-86bed846ff53cc9ec2c51bcb085621d4R48) anyway. Let me know if I'm missing something here.

awolfe76:
@wpears - this is ready, but be sure to check out https://github.com/cfpb/hmda-platform-ui/pull/745/files#r137295513.

src/js/containers/Institutions.jsx
wpears: The submission exists once it's been created and once the submission page is visited, but should always result in the same value as this method.

That said, this method requires that submissions are returned in reverse sorted order (which is currently true, but is undocumented and subject to change).

I'd be more comfortable with a function that takes the submissions array, loops over it, and returns the one with the highest `sequenceNumber`

awolfe76: I just brought it back, for now, was easier to do (78511bd).

awolfe76:
@wpears 78511bd brought back the `submission`. I think if we want to do even more refactoring here we can create issues for it, eg for the looping over filings to set the submission.

wpears (approved):

PR #746, opened 2017-08-31
update language for modals

### LINKED ISSUE #675: Wording for modals
We have 2 modals. The design is consistent, but _I think_ the content could be better. Rather than the title for both being "HMDA Filing" the titles should be more about what the modal is doing. Something like "You have timed out." and "You are about to refile." and then have an appropriate message/description in the body along with the buttons.

The goal of this issue is to decide on both the titles and the body for both modals.

# Current modals

## Refile
modal-refile

## Timeout
modal-timeout

awolfe76:
Closes #675.


wpears:
:+1:

PR #761, opened 2017-09-08
Sorting

### LINKED ISSUE #750: Sort institutions when multiple institutions exist
When a user has multiple institutions, say Bank0 and Bank1, they can be sorted differently when going to `/institutions`. We should make this consistent, maybe sorting by the `institution id`.

awolfe76:
Closes #750

### DIFF
--- a/src/js/containers/Institutions.jsx
+++ b/src/js/containers/Institutions.jsx
@@ -3,7 +3,6 @@ import { connect } from 'react-redux'
 import fetchInstitutions from '../actions/fetchInstitutions.js'
 import fetchCSV from '../actions/fetchCSV.js'
 import Institutions from '../components/Institutions.jsx'
-import sortFilings from '../utils/sortFilings.js'
 
 export class InstitutionContainer extends Component {
   constructor(props) {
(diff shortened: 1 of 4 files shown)


src/js/containers/Institutions.jsx
```diff
@@ -21,8 +22,11 @@ export class InstitutionContainer extends Component {
 
 export function mapStateToProps(state) {
   const { institutions } = state.app.institutions
+  const { submission, error, filingPeriod } = state.app
+  const unsortedFilings = state.app.filings
 
-  const { filings, submission, error, filingPeriod } = state.app
+  let filings = unsortedFilings
+  filings.filings = unsortedFilings.filings.sort(sortFilings)
```
wpears: I don't think we should mutate the state here, was the sort not working in the `receiveInstitutions` action creator?

wpears (changes requested):
One above

src/js/containers/Institutions.jsx
awolfe76: 👍  I'll keep you posted.

PR #763, opened 2017-09-11
Update wording to lend a greater sense of completion to signature message

### LINKED ISSUE #755: Crank up the signature message happiness even more
https://github.com/cfpb/hmda-platform-ui/pull/747 previously updated the success message, but we need more voltage there to ensure the user knows they are done (and that they're getting an email saying as much as well, see https://github.com/cfpb/hmda-platform/issues/1126).

There should be a line after the header reading something like:
Congratulations, you've successfully completed your HMDA filing for `<filing period>`!

wpears:
*ie* signature happiness
Closes #755 
screen shot 2017-09-11 at 10 13 07 am

### DIFF
--- a/src/js/components/Signature.jsx
+++ b/src/js/components/Signature.jsx
@@ -12,13 +12,10 @@ const showReceipt = ({status, timestamp, receipt, filingPeriod, email}) => {
   return (
     <Alert type="success" heading="HMDA filing completed">
       <div>
-        <hr/>
         <p>Congratulations, you have successfully completed your HMDA filing for {filingPeriod}!</p>
-        <p>Your data and signature have been received and recorded.</p>
-        <p className="usa-text-small">Your submission was received on <strong>{ordinalHour(new Date(timestamp))}</strong>.</p>
-        <p className="usa-text-small">Your receipt number is <strong>{receipt}</strong>.</p>
-        <hr/>
-        <p>You will receive a confirmation email with this information shortly at <strong>{email}</strong>.</p>
+        <p>Your data and signature were received and recorded on <strong>{ordinalHour(new Date(timestamp))}</strong>.</p>
+        <p>Your receipt number for this submission is <strong>{receipt}</strong>.</p>
+        <p>You will receive a confirmation email shortly at <strong>{email}.</strong></p>
       </div>
     </Alert>
   )
--- a/src/js/components/Signature.jsx
+++ b/src/js/components/Signature.jsx
@@ -5,14 +5,20 @@ import { ordinalHour } from '../utils/date.js'
 import Alert from './Alert.jsx'
 import { VALIDATED_WITH_ERRORS, SIGNED } from '../constants/statusCodes.js'
 
-const showReceipt = (code, timestamp, receipt) => {
+const showReceipt = ({status, timestamp, receipt, filingPeriod, email}) => {
+  const code = status.code
   if (code !== SIGNED) return null
 
   return (
-    <Alert type="success" heading="HMDA data accepted">
+    <Alert type="success" heading="HMDA filing completed">
       <div>
-        <p>Your submission was received on <strong>{ordinalHour(new Date(timestamp))}</strong>.</p>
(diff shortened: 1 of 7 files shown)


awolfe76 (commented):
LGTM, something does seem off about the multiple font sizes, feels a bit scattered that way.

But would like for @designlanguage and @keileyo to make sure the wording is 👍 .

src/js/components/Signature.jsx
```diff
@@ -5,14 +5,20 @@ import { ordinalHour } from '../utils/date.js'
...
+  const code = status.code
   if (code !== SIGNED) return null
 
   return (
-    <Alert type="success" heading="HMDA data accepted">
+    <Alert type="success" heading="HMDA filing completed">
       <div>
-        <p>Your submission was received on <strong>{ordinalHour(new Date(timestamp))}</strong>.</p>
-        <p>Your receipt number is <strong>{receipt}</strong>.</p>
+        <hr/>
```
awolfe76: I think we'd be ok without the `<hr>` tags in this alert, including the one below. We haven't used them in other alerts and there's not a real change in content here. The messaging is all about filing completion.

src/js/components/Signature.jsx
```diff
@@ -5,14 +5,20 @@ import { ordinalHour } from '../utils/date.js'
...
+        <p>Congratulations, you have successfully completed your HMDA filing for {filingPeriod}!</p>
+        <p>Your data and signature have been received and recorded.</p>
+        <p className="usa-text-small">Your submission was received on <strong>{ordinalHour(new Date(timestamp))}</strong>.</p>
```
awolfe76: I think we should remove the `usa-text-small` class here. The font size jump (from heading to p to small back to p) makes it feel like there's too much going on.

src/js/components/Signature.jsx
```diff
@@ -5,14 +5,20 @@ import { ordinalHour } from '../utils/date.js'
...
   if (code !== SIGNED) return null
 
   return (
-    <Alert type="success" heading="HMDA data accepted">
+    <Alert type="success" heading="HMDA filing completed">
       <div>
-        <p>Your submission was received on <strong>{ordinalHour(new Date(timestamp))}</strong>.</p>
-        <p>Your receipt number is <strong>{receipt}</strong>.</p>
+        <hr/>
+        <p>Congratulations, you have successfully completed your HMDA filing for {filingPeriod}!</p>
```
awolfe76: I think we can combine theses paragraph tags making it `<p>Congratulations, you have successfully completed your HMDA filing for {filingPeriod}! Your data and signature have been received and recorded.</p>`.

There doesn't seem like a big need for them to be broken out.

src/js/components/Signature.jsx
awolfe76: Same as above here, what if we combine these to be `<p>Your submission was received on <strong>{ordinalHour(new Date(timestamp))}</strong> and Your receipt number is <strong>{receipt}</strong>.</p>`.

awolfe76 (changes requested):
A few things here just to make it feel less "scattered".

#### Example

signature

wpears:
@designlanguage Thoughts on the above?
The original reason I separated each line was to give each statement more visual weight and to provide the traditional line-item appearance of the receipt.
The horizontal rules were provided for structure and emphasis and the font size changes were to add a bit of hierarchy (normal size: things you should read/know, usa-text-small: errata that may be useful but isn't really essential to read at this time).

designlanguage:
@wpears - I think we can compress this info a bit.

The message seems to be made up of three parts: What, When, and Where. I'm interested to know if any of these are less important, but it seems like more of a progression to me, so I'm for keeping them the same size. If we start a new line for each it could look like this:

### ✔  __HMDA filing submitted__

Congratulations, you successfully completed your HMDA filing for 2017!

Your data and signature were received and recorded on __September 12th, 2017, 1:28:14 pm.__

You will receive an email confirmation shortly at __NameATBank0.com__. Your receipt number for this submission is __00000000-0000000-0000000000000__.

awolfe76 (approved):

PR #764, opened 2017-09-12
Confirmation review

### LINKED ISSUE #756: Note on confirmation page to review and sign
Some people thought that they were done upon reached the confirmation page, a green-background `alert-success` box with something like:
```
Your filing is ready to be signed and submitted, please review your filing summary and sign your filing at the bottom of this page.

If you discover an error in the summary you will need to update your file and upload it again.
```

Some messaging can then be pulled out of the IRS component (which, incidentally, has a typo https://github.com/cfpb/hmda-platform-ui/blob/master/src/js/components/IRSReport.jsx#L154 Update a new file -> Upload a new file)

wpears:
Renders an info-level alert on the confirmation page to let people know they need to sign below.
Depends on #763 
Closes #756


src/js/components/Alert.jsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 
-const Alert = ({ type = 'info', heading, children }) => {
+const Alert = ({ type = 'info', heading, linebreak, children }) => {
```
awolfe76: Same as the last PR, I don't think we need the break (`<hr />`) here. The content is good without it.

awolfe76 (changes requested):
Feels like this alert might be better treated the same way we do the alerts on the edits pages (right at the top and 100% width. And also with a heading, the `Your filing is ready to be signed and submitted.` line could be the heading. Could treat the new component the same way we did the `<RefileWarning />` [here](https://github.com/cfpb/hmda-platform-ui/blob/3a68b61ab1755139ba6eb673e91b68593a63dd0f/src/js/containers/Submission.jsx#L116).

src/js/components/Alert.jsx
wpears: Yeah, this was actually just a reorg of code from that first PR anyhow, knew there was going to be more changes here once that was merged ^^

wpears:
I thought that might be the case, makes sense to me.

wpears:
Should be good to go @awolfe76

PR #765, opened 2017-09-12
Confirmation to submission

### LINKED ISSUE #758: Rename step 5?
Based off what testing revealed about the confirmation page, maybe we should change its name.

How about we label Step 5 as "Submission" instead?

@keileyo

wpears:
Depends on #764 
Closes #758


src/js/components/Signature.jsx
```diff
@@ -10,7 +10,7 @@ const showReceipt = ({status, timestamp, receipt, filingPeriod, email}) => {
   if (code !== SIGNED) return null
 
   return (
-    <Alert type="success" heading="HMDA filing completed">
+    <Alert type="success" heading="HMDA filing completed" linebreak={true}>
```
awolfe76: We can remove the `linebreak={true}`.

src/sass/app.scss
```diff
@@ -14,6 +14,7 @@
 
 // Components
 @import "components/Alert.scss";
+@import "components/SubmissionPageInfo.scss";
```
awolfe76: We aren't using this right? That component takes the [`RefileWarning` styles](https://github.com/cfpb/hmda-platform-ui/blob/0d0cf9fdb65886ba6a215bc8bee54fb8e3138f15/src/js/components/SubmissionPageInfo.jsx#L6).

src/sass/components/SubmissionPageInfo.scss
```diff
@@ -0,0 +1,5 @@
+.SubmissionPageInfo {
```
awolfe76: We can get rid of this file.

awolfe76 (changes requested):
Just a couple quick things.

wpears:
Yuck, that stuff came back in when I rebased, removed now.

awolfe76 (approved):
Nice!

PR #766, opened 2017-09-12
add logo to header

### LINKED ISSUE #754: Branding / official gov site
Some users wanted more than our banner (or maybe didn't notice it). We could add the [updated banner](https://www.usa.gov/) to have a bit more explanation here.

We're kinda in a bind due to the requirement of not having our logo in the top left (which is kinda the official hey-this-is-who-this-app-is-associated-with sign).

Ultimately, we can probably skate by here knowing that *usually* people will only be directed here from other more branded sites/entities.

awolfe76:
Closes #754 

This adds the logo to the header but also removes the link from the "HDMA Platform" text. This link took the filers to the home page. We did add a "HOME" link to the navigation a while back to take care of this.

Also in that thought was that the footer still has the FFIEC logo and it links to the [FFIEC home page](https://www.ffiec.gov/). It felt weird that the header logo/platform text would go to the app home page but the footer logo would go somewhere else. _Screenshots are below to show where we are using the logo._

Should we also remove the link from the FFIEC logo in the footer as well (we have another link in the footer that goes there)? Other thoughts?

_Also some formatting changes, [here are the important lines](https://github.com/cfpb/hmda-platform-ui/pull/766/files?w=1#diff-92687e6d8308aa62dd8f6bfd53b414d3R63)._

### Header
header

### Footer
footer

### DIFF
--- a/src/js/components/Footer.jsx
+++ b/src/js/components/Footer.jsx
@@ -1,0 +1,74 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { Link } from 'react-router'
+
+const Footer = () => {
+  return (
+    <footer className="Footer usa-footer usa-footer-slim" role="contentinfo">
+      <div className="usa-grid usa-footer-return-to-top">
+        <a href="#">Return to top</a>
+      </div>
+      <div className="usa-footer-primary-section">
+        <div className="usa-grid-full">
+          <nav className="usa-footer-nav usa-width-one-third">
+            <ul className="usa-unstyled-list">
+              <li className="usa-footer-primary-content">
+                <Link
+                  className="usa-nav-link"
+                  to={'/'}
+                  title="Home"
+                  aria-label="Home"
+                >
+                  <img src="/img/ffiec-logo.png" width="75px" alt="FFIEC" />
+                  HMDA Platform
+                </Link>
+              </li>
+            </ul>
+          </nav>
+          <div className="usa-width-one-third">
+            <h4>Resources</h4>
+
+            <ul className="usa-unstyled-list">
+              <li>
+                <a href="https://www.ffiec.gov/hmda/">FFIEC HMDA Website</a>
+              </li>
+              <li>
+                <a href="https://www.federalregister.gov/documents/2015/10/28/2015-26607/home-mortgage-disclosure-regulation-c">
+                  HMDA Final Rule
+                </a>
+              </li>
+              <li>
+                <a href="https://www.consumerfinance.gov/policy-compliance/guidance/implementation-guidance/hmda-implementation/">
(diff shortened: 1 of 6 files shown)


wpears (changes requested):
So I think if we're including the FFIEC logo in the header we **need** it to link to ffiec.gov.

Also, I think the hierarchy is a bit off here, the large bold, dark characters of HMDA Platform overpower the FFIEC logo. Try setting the width of the image to 140px, margin-top to 1em and margin-bottom to -0.15em for something that I think fits better.

awolfe76:
@wpears - I agree, but instead I shrunk the `font-size` to bring close to the same size as the logo "text"  and the `font-weight` so the text isn't nearly as strong.

> So I think if we're including the FFIEC logo in the header we need it to link to ffiec.gov.

I see what you mean, but I also see potential there for a filer to be confused. The location (top left) of the logo is a common location on web apps/sites for a link to the app/site home page, having it jump to a whole other site seems confusing to me and I don't think it's likely that a filer would need, or want, to visit that site while they are filing.

wpears:
I'd say if we're not willing to link to the site, we shouldn't include the logo in the header at all. Or perhaps, we wrap the logo and the HMDA Platform text in a link which just goes to the platform homepage. I think removing the link altogether though takes away more than it gives.

Thoughts @designlanguage @keileyo

designlanguage:
@awolfe76 - I don't think the FFIEC logo in the header or footer should link to anything, exactly because of what you said. I think the logo should remain in the footer though.

Also, I don't necessarily agree that the link from HMDA Platform should be removed though. Maybe there's no underline, but I don't see any harm in having that be another way to get back to the home page. Anyone clicking on it wouldn't be trying to do anything else.

awolfe76:
I'm ok linking both the home page of the platform. Would still like to hear @designlanguage and @keileyo? The idea here, from testing, was that users wanted it to be a bit more clear they were on a `.gov`.

designlanguage:
I didn't see @wpears's comment before I posted.  I think keeping the HMDA Platform text always locked up with the FFIEC logo in the top and bottom, and linking both to the home page makes sense to me.

awolfe76:
Will work on [this](https://github.com/cfpb/hmda-platform-ui/pull/766#issuecomment-329229255).

awolfe76:
@wpears - 6383d93 and fd5633e take care of adding the header and footer links. I also pulled the footer out into its own component.

wpears:
Before we merge this, can we get an implementation for https://github.com/cfpb/hmda-platform-auth? Both in the header and the footer. Looks good otherwise 👍

awolfe76:
Good call @wpears - https://github.com/cfpb/hmda-platform-auth/pull/162

PR #767, opened 2017-09-12
Set envvars for non-default docker-machine host

charandas:
I noticed that the docker-compose file in hmda-platform does not provide the `_SERVER` env vars leading to this not working for non-default docker-machine hosts. I am running the platform with "Docker for mac/Moby".

Extracting the base url from `_URL` vars seems to make sense for HMDA_API_URL and KEYCLOAK_URL. Let me know if I can improve the PR further.

### DIFF
--- a/env.sh
+++ b/env.sh
@@ -1,5 +1,14 @@
 #!/bin/bash
 
+get_baseurl () {
+  local proto=$(echo ${1} | sed -e's,^\(.*://\).*,\1,g')
+  local baseUrl=$(echo ${1} | awk -F/ '{print $3}')
+  echo "${proto}${baseUrl}"
+}
+
+KEYCLOAK_SERVER=$(get_baseurl ${KEYCLOAK_URL})
+HMDA_API_SERVER=$(get_baseurl ${HMDA_API})
+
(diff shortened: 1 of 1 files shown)


env.sh
```diff
@@ -1,5 +1,14 @@
 #!/bin/bash
 
+get_baseurl () {
+  local proto=$(echo ${1} | grep :// | sed -e's,^\(.*://\).*,\1,g')
```
wpears: What's the function of the grep here? Echoing straight into the sed would work just as well afaik.

wpears (changes requested):
Hey thanks for your work here, the duplication of the env_vars was pretty hacky and I've been meaning to revisit it.

I like this solution specifically because the variables *should* be derived, since their use in the CSP would preclude them from being separate servers.

Could you trim out the grep for the protocol as mentioned below? Looks good otherwise

env.sh
charandas: Good point. I will get rid of the grep.

wpears (approved):

wpears:
Cool, thanks!

PR #768, opened 2017-09-12
Sort filings (institutions)

### LINKED ISSUE #750: Sort institutions when multiple institutions exist

awolfe76:
Closes #750 

#### `699999` is orignally first, but 2nd in the 2nd (sorted) output

sorted


wpears (approved):
👍

PR #769, opened 2017-09-13
let the user know that they can logout

### LINKED ISSUE #724: Add language to let filer know they can logout
It would be nice during a long validation process to message to the filer that they can leave/logout and come back later. See #642 and #717 for more details.

awolfe76:
Closes #724 

I left this dependent on the large file, so if a filer uploads a large filer and the file status is > uploaded and hasn't completed validation and didn't have parsing errors, they will see a message saying 

`Your upload will complete automatically, so you may leave the platform and log back in later.`

#### Screenshot

screen shot 2017-09-13 at 9 15 22 am

### DIFF
--- a/src/js/components/ValidationProgress.jsx
+++ b/src/js/components/ValidationProgress.jsx
@@ -34,14 +34,14 @@ export default class ValidationProgress extends Component {
   }
 
   getText() {
-    let text = 'Uploading...'
+    let text = 'Uploading ...'
     const code = this.props.code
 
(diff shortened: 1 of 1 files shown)


wpears:
I think we should still keep in the "This may take several minutes" bit to set expectations, otherwise it seems strange just telling people they can leave.

awolfe76:
@wpears see e8fa3b0. I said "a little while" instead of "several minutes". Minutes may always apply but I'm not sure if we know enough about timing yet, especially in the case of large files.

PR #772, opened 2017-09-13
Polling backoff

### LINKED ISSUE #770: Add exponential backoff to polling
For large files, polling continuously at a 1s interval is a clear waste of resources.

We should enable an exponential backoff (with some hard cap) so that, at the cost of a slightly delayed upload we can reduce server load significantly.

wpears:
Closes #770 

During upload, the `pollForProgress` action now slowly backs off, up to a maximum of a request every 20s. This will greatly reduce network traffic to the backend with a fairly small delay relative to total upload time.

### DIFF
--- a/src/js/actions/pollForProgress.js
+++ b/src/js/actions/pollForProgress.js
@@ -5,7 +5,19 @@ import hasHttpError from './hasHttpError.js'
 import { getLatestSubmission } from '../api/api.js'
 import { PARSED_WITH_ERRORS, VALIDATED_WITH_ERRORS } from '../constants/statusCodes.js'
 
+export const makeDurationGetter = () => {
+  let count = 0
+  return () => {
+    let duration = Math.pow(1.2, count)*1000>>0
+    if (duration > 30000) duration = 30000
+    else count++
+    return duration
+  }
+}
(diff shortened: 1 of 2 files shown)


src/js/actions/pollForProgress.js
```diff
@@ -22,7 +34,7 @@ export default function pollForProgress(polling) {
       .then(json => {
         if(json.status.code < VALIDATED_WITH_ERRORS &&
            json.status.code !== PARSED_WITH_ERRORS){
-             setTimeout(() => poller(dispatch), 1000)
+             setTimeout(() => poller(dispatch), getTimeoutDuration())
```
awolfe76, line 37: Just a quick question, why use `getTimeoutDuration()` here instead of just `makeDurationGetter()`?

wpears: it would have to be `makeDurationGetter()()` since it returns a function.

I designed it this way so that the duration getting was testable, on my first pass I just defined `getTimeoutDuration` in the callback, but then it was inaccessible.

awolfe76: 👍

PR #784, opened 2017-09-20
Institution error

### LINKED ISSUE #730: No filings on an institution
### Current

RIght now it renders like this 

no-filings

### Proposal

We should still show the filer the institution name and other information. It should closely resemble the same as having a filing. Something like

no-filings-better

This would let the filer see which institution has the problem, especially helpful when they can file for more the 1.

awolfe76:
Closes #730 

Refactored to loop over `institutions` instead of `filings.filings` to allow for better error messages and to allow an error to be associated with an institution.

### Screenshot with multiple institutions with no filings

no-filings


wpears:
:+1:

PR #786, opened 2017-09-25
Don't embed css sourcemap

wpears:
Greatly reduces css file size


awolfe76 (approved):
I lose the source-map to debug after a rebuild when running in the container, but I can see that its (re)created, so it's a different problem. 👍

PR #791, opened 2017-09-26
Bug upload new

### LINKED ISSUE #783: Upload a new file link broken with multiple institutions
When a filer has multiple institutions listed _sometimes(?)_ the 'upload a new file' link will sometimes create a new submission for the wrong institution.

A possible fix to this is to reset the `institution` when the `/institution` route is taken ... so having a `clearInstitution` action distpatched [around here](https://github.com/cfpb/hmda-platform-ui/blob/8dc6bd9d9660f0aab863bfcbea1900a43ba7f654/src/js/containers/Institutions.jsx#L12) could do the trick.

It's safe to use the `state.app.institution` on other routes because the institution will be correctly set. This is __not__ the case on the `/institutions` route.

awolfe76:
Closes #783


awolfe76:
@wpears - this should be good now. 4728bd6 fixed a bug I found.

wpears:
👍

Issue #657, opened 2017-07-13
Ensure bundle-size is as small as possible


wpears:
- The jsrsasign is the main culprit by way of oidc-client
   - This may be able to be modularized by pulling out necessary code and aliasing requires
 - Moment can be replaced with a small module of custom code
 - Preact can replace react/react-dom, but probably shouldn't be done before beta release
 - Babel-polyfill is reasonably large, but we need it for some dependencies (oidc-client, among others)
    - This includes our own code, but we *could* isolate what we actually need and include it modularly from core-js, rather than including the entire core-js package as part of babel-polyfill

Issue #789, opened 2017-09-25
Remove nodemon dependency


awolfe76:
Nodemon is buggy.

There is a `--watch` option for both [jest](http://facebook.github.io/jest/docs/en/cli.html#watch) and [node-sass](https://github.com/sass/node-sass#usage-1).

Browserify can be configured a few ways but [watchify](https://github.com/browserify/watchify) seems like a good option.

awolfe76:
Closed by #809.

PR #793, opened 2017-09-26
Switch to Webpack

### LINKED ISSUE #788: Switch to babel-preset-env
This would allow us to better target browsers, https://github.com/babel/babel-preset-env.

This should *__not__* require a CR because its a dev dependency.

wpears:
Closes #788 
Will address #789 separately


awolfe76:
@wpears - can we close #657 with this one too? And maybe open issues when we find stuff later?

.babelrc
```diff
@@ -1,7 +1,19 @@
 {
   "presets": [
     "react",
-    "es2015",
+    ["env", {
+      "targets": {
+        "useBuiltIns": "entry",
+        "browsers": [
```
awolfe76, line 7: Since we use this [here](https://github.com/cfpb/hmda-platform-ui/pull/793/files#diff-11e9f7f953edc64ba14b0cc350ae7b9dR39) and [here](https://github.com/cfpb/hmda-platform-ui/pull/793/files#diff-5b57b65a004a69d9237f08df4fc77e53R39), should we pull it out into a separate file to make it easier to update?

awolfe76 (changes requested):
Everything looks good, but can we add the source-map back?

.babelrc
wpears: I'm doing this in the nodemon PR, creating a common config that the dev and prod setups inherit

wpears:
I want to keep #657 open still, have some more ideas there

awolfe76 (approved):

PR #804, opened 2017-09-28
Update dockerfile to centos and update readme

### LINKED ISSUE #787: Upgrade Yarn
### This will require a CR.

When we have a chance we should [upgrade our Yarn version](https://github.com/cfpb/hmda-platform-ui/blob/76e429efbfce9d3fba5b42e35e670b87fad09c9b/Dockerfile#L10). They have released [1.0.0 and beyond](https://github.com/yarnpkg/yarn/releases) with performance gains.

### LINKED ISSUE #798: Switch to Centos-based image
Better mirrors platform auth and removes dev dependency on the debian-based nginx image

wpears:
Closes #798 
Closes #787

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,25 +1,28 @@
-FROM nginx:1.10
-MAINTAINER Wyatt Pearsall<Wyatt.Pearsall@cfpb.gov>
-ARG SKIP_JS_BUILD
+FROM centos:7
 
-RUN if [ -z ${SKIP_JS_BUILD+x} ]; then echo "Installing JS deps" && apt-get update && \
-    apt-get install -y curl && \
-    curl -sL https://deb.nodesource.com/setup_8.x | bash - && \
-    apt-get install -y nodejs && \
-    curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && \
-    echo "deb https://dl.yarnpkg.com/debian/ v0.27.5 main" | tee /etc/apt/sources.list.d/yarn.list && \
-    apt-get update && \
-    apt-get install -y yarn; fi && \
-    mkdir -p /usr/src/app && \
-    chown -R root /usr/src/app && chmod u+rwx /usr/src/app
+ENV NGINX_USER=svc_nginx_hmda
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -28,7 +28,7 @@ _All pages, expect the home page, require authentication._
 
 ## Dependencies
 
-We use npm and yarn to manage front-end dependencies. You should have [npm installed](https://nodejs.org/en/) as well as [yarn](https://yarnpkg.com/lang/en/docs/install/).
+We use node and yarn to manage front-end dependencies. You should have [node installed](https://nodejs.org/en/) as well as [yarn](https://yarnpkg.com/lang/en/docs/install/).
```
awolfe76, line 31: Should we update `package.json` too, to remove `npm` from the scripts?

awolfe76 (changes requested):
Just the one thing about removing `npm` completely. Otherwise looks good.

README.md
wpears: I'll do that as part of #789

awolfe76: 👍

Dockerfile
```diff
@@ -1,25 +1,28 @@
...
+    yum-config-manager --enable cr && \
+    yum update -y && \
+    yum install -y nginx-1.10.2 \
+    yum clean all && \
+    usermod -l $NGINX_USER nginx && \
+    groupmod -n $NGINX_USER nginx && \
+    rm -rf /usr/share/nginx/ && \
+    ls -d -1 /etc/nginx/* | grep -v '\/mime.types$' | xargs rm -rf && \
+    mv nginx/* /etc/nginx && \
+    sed -i 's/80/8080/g' /etc/nginx/nginx.tmpl && \
```
hkeeler, line 19: Sorry, just now checking this out.  Now that we're running the container on 8080 here, might as well tweak the [`nginx.tmpl`](https://github.com/cfpb/hmda-platform-ui/blob/master/nginx/nginx.tmpl#L41) to also have 8080, then we won't have to do this find/replace here.  :)

wpears: Roger

PR #805, opened 2017-09-28
Submission page message

### LINKED ISSUE #795: Remove "ready to be signed" banner once signed

wpears:
Closes #795 

Also, adds the success message/receipt as an alert header on the submission page once signed so return visitors get instant feedback.

### DIFF
--- a/src/js/containers/Submission.jsx
+++ b/src/js/containers/Submission.jsx
@@ -25,7 +25,8 @@ import {
   UNINITIALIZED,
   FAILED,
   PARSED_WITH_ERRORS,
-  VALIDATING
+  VALIDATING,
+  SIGNED
 } from '../constants/statusCodes.js'
 
 const Edits = submissionProgressHOC(EditsContainer)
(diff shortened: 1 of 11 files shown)


src/js/containers/Submission.jsx
```diff
@@ -10,6 +10,7 @@ import SubmissionPageInfo from '../components/SubmissionPageInfo.jsx'
 import UploadForm from './UploadForm.jsx'
 import ErrorWarning from '../components/ErrorWarning.jsx'
 import EditsContainer from './Edits.jsx'
+import Receipt from './Receipt.jsx'
```
awolfe76, line 13: It's picky, but I think we should rename this `SubmissionReceipt`. At least it better matches `SubmissionPageInfo` and that it falls on the submission page.

_Which brings up the directories again._ 😉

awolfe76 (changes requested):
Just the one small thing.

awolfe76 (approved):

PR #807, opened 2017-09-29
Switch nginx port and update Dockerfile accordingly

wpears:
We were doing an unnecessary sed now that the dockerfiles are unified.

Note: bugfix for failing test is in #808


hkeeler (approved):
👍

PR #809, opened 2017-09-29
Remove nodemon

wpears:
Currently without yarn on my local setup (long story), can you test this @awolfe76? All the commands seem to work as expected with `npm`. Jest's watch flag is a little wonky/blocking so I left that out of the `watch` command, recommend `watch` and `watch:tests` be run in separate terminal panes/instances.

Also, you'll need to add some packages to the offline cache in a commit to make travis happy (since we run offline in travis).
Steps:
Delete `node_modules`
Delete `yarn.lock`
cd npm-packages-offline-cache and `rm *`
Edit .yarnrc and add comments before the two flags, ie
```
#--install.pure-lockfile true
#--install.offline true
```
Run `yarn`

This should update everything appropriately, then remove the comments from .yarnrc and commit everything.


awolfe76:
@wpears - added a few small commits and the [one comment/question](https://github.com/cfpb/hmda-platform-ui/pull/809/files#r141884419).

4411c6f creates a `js:dev` task and config to take dev build times from ~25 seconds to ~10 seconds.

Also, Travis is still failing with `Can't make a request in offline mode`. I'm looking into it but if you know what's up that would be 🚀 .

package.json
```diff
@@ -9,36 +9,28 @@
...
+    "js": "webpack --progress",
+    "js:dev": "webpack --config webpack.dev.js --progress",
+    "renew": "webpack --config webpack.renew.js --progress",
+    "sass": "node-sass --quiet --output-style=compressed --source-map=./dist/css/ ./src/sass/app.scss ./dist/css/app.min.css && cp node_modules/uswds/src/fonts/sourcesanspro-* ./dist/fonts",
+    "html": "html-minifier --collapse-whitespace ./src/index.html -o ./dist/index.html && html-minifier --collapse-whitespace ./src/silent_renew.html -o ./dist/silent_renew.html",
+    "img": "cp ./src/img/min/* ./dist/img && cp -r ./node_modules/uswds/dist/img/* ./dist/img",
+    "postinstall": "yarn clean && yarn build && yarn renew",
+    "watch:js": "yarn js:dev -- --watch",
+    "watch:sass": "node-sass --watch --quiet --output-style=compressed --source-map=./dist/css/ ./src/sass/app.scss ./dist/css/app.min.css",
+    "watch:tests": "jest --watchAll",
```
awolfe76: What do you think about using [`--watch`](https://facebook.github.io/jest/docs/en/cli.html#watch) instead? To get the full report we still have the `test` script.

wpears: Just --watch is fine, I found it to not always be 100% accurate, but it's probably good enough in most situations. Also, if other tests fail that it doesn't find, that's probably an issue with our test architecture/tests being too coupled.

awolfe76: To be honest I'm fine either way. I usually just use `jest --watch` anyway. :) But this is still convenient to have here.

wpears:
@awolfe76 Sass woes should be fixed with the latest

src/sass/components/_Home.scss
```diff
@@ -1,6 +1,6 @@
 .Home {
   .usa-grid {
-    margin-top: 2em;
+    margin-top: 20em;
```
awolfe76: I don't think you meant to commit this change. 😄

src/sass/components/_Alert.scss
```diff
@@ -6,7 +6,7 @@
     margin-top: -1em;
     margin-bottom: -0.3em;
     height: 30px;
-    width: 30px;
+    width: 10px;
```
awolfe76: I don't think you meant to commit this change. 😄

package.json
```diff
@@ -47,27 +40,29 @@
   "homepage": "https://github.com/cfpb/hmda-platform-ui#readme",
   "devDependencies": {
     "babel-jest": "20.0.3",
-    "babel-loader": "^7.1.2",
+    "babel-loader": "7.1.2",
     "babel-plugin-transform-object-rest-spread": "6.19.0",
     "babel-polyfill": "6.16.0",
     "babel-preset-env": "1.6.0",
     "babel-preset-react": "6.16.0",
+    "core-js": "2.5.1",
```
awolfe76: Is this being used anywhere?

_While we're at it, I think we can remove `synk`, `mkdirp`, `immutable`, and `minifyify`._
_And `react-router-scroll` should probably be a dependency not a devDependency._

awolfe76 (changes requested):
Just a few commits that I don't think you meant to make and a little clean up on `package.json`.

awolfe76 (approved):
💯

PR #812, opened 2017-10-02
Render loader when irs is fetching

### LINKED ISSUE #801: IRS loading icon
The IRS report could take a while to return. We should add a loading icon to it.

It looks like we already have the [`prop`](https://github.com/cfpb/hmda-platform-ui/blob/f93f490133c30c6031ef2d962b032eb4d0521fe1/src/js/components/IRSReport.jsx#L172), just need to use it.

wpears:
Closes #801


awolfe76 (approved):
👍

PR #813, opened 2017-10-04
Allow Google tag manager and Google analytics to work

### LINKED ISSUE #803: Google Tag Manager 404
We're getting a 404 for GTM (in auth too). Need to investigate and test this in production.

awolfe76:
Closes #803


awolfe76:
This should be ready @wpears. To see the changes, other than formatting, in `index.js` check 
edf0c8e and 63f6681.

I'm also not tracking on `/oidc-callback`. If you think we should let me know.

I've created #824 to get more detailed but I think this is a good start.

PR #828, opened 2017-10-06
apply prettier

awolfe76:
@wpears - this is pretty big. All `yarn` commands and smoke tests still seem to be 💯. But definitely take a look. I'd like to take a closer look myself before merge.

The [constants file](https://github.com/cfpb/hmda-platform-ui/pull/828/files#diff-9d96c3b57d7ab20eff30db9601e41a1d) for one. I could go either way here but it's worth seeing.

And make sure to check my config and the README.


awolfe76:
@wpears - I can split this up a bit too if you'd like?

src/js/components/EditsNav.jsx
```diff
@@ -19,11 +19,11 @@ const navNames = [
 ]
 
 const navLinks = {
-  'upload': 'upload',
+  upload: 'upload',
```
awolfe76, line 22: This is another sort of odd one, but it's ok with me.

wpears:
Cool stuff, no more wonky formatting in commits 🗡

PR #833, opened 2017-10-13
Remove cluttered output and fixup testing errors

wpears:
Closes #827

PR #834, opened 2017-10-13
Registration link

### LINKED ISSUE #821: Signup link in logged out home page
Users had some confusion about where they would go to sign up. They eventually selected the Begin filing or Login button, but they didn't feel either described what they wanted to do. I think we can limit hesitation/friction in the account creation process one of two ways:

* We could include a text link next to the login button that says: "__Create an account__" that goes directly to the account creation page.

* Alternatively, we could add a link in the disclaimer text below the button where it says "register online", but the text is pretty tiny at the moment. If we go with this option, I think the disclaimer text would need to be larger.

screen shot 2017-10-05 at 9 00 25 am

wpears:
Closes #821 
Depends on #833 

Adds registration link to signed out homepage, overriding the getAuthorizationEndpoint method of oidc-js.


awolfe76 (approved):
👍

PR #851, opened 2017-10-25
File errors

### LINKED ISSUE #850: Invalid File Format
This is the message that displays if you upload a non-txt file:

image

If you return to the home page, you will see this. After clicking on view current filing, you will get a blank screen (second image below). The only way to restart the filing process is to log out and log back in.
image
image

wpears:
Closes #850


awolfe76:
@wpears - if you
- upload a good file, `good-file.txt` and it runs
- then upload a `bad-file.dat` you'll see the message in the dropzone says `good-file.txt can not be uploaded.`

That should change to the `bad-file.dat` filename.

src/js/components/UploadForm.jsx
```diff
@@ -152,9 +146,6 @@ export default class Upload extends Component {
 }
 
 Upload.propTypes = {
-  setFile: PropTypes.func,
-  setNewFile: PropTypes.func,
-  showConfirmModal: PropTypes.func,
   pollSubmission: PropTypes.func,
```
awolfe76: we should add a `handleDrop` prop type here.

wpears:
@awolfe76 should be good to go now

PR #856, opened 2017-10-30
Fix multi institution routing and add tests

### LINKED ISSUE #854: Switching between multiple institutions fails
If you have more than one institution trying to go from viewing a submission in an institution, then going back to the home page and trying to view another, you will see only the loading icon.

loading-icon

wpears:
Closes #854


awolfe76 (approved):
👍

PR #866, opened 2017-11-02
Warnings

### LINKED ISSUE #829: Duplicate alert banner functionality at the bottom of steps 2, 3, 4
Since the alerts aren't persistent (to preserve vertical space), the user won't see the call to action after they've scrolled. We should have those same actions at the bottom of each edits page, especially if there are a lot of edits to review.

### S/V
The strongest case for this is the Syntactical/Validity page. If there are edits here, there is no action that can be taken once a filer has scrolled to the bottom of the page. When S/V edits exist, I propose including modified banner content at the bottom. When they don't exist, it would not be present.

| Current | Proposed |
| --- | --- |
| screen shot 2017-10-10 at 11 36 50 am | screen shot 2017-10-10 at 11 35 17 am |

The content could say something like: "Please review the syntactical and/or validity edits above or __Download the edit report__. Then update your file and select the “Upload a new file” button."

### Q and MQ

For quality and macro, this might be a little different. We already have the verify action at the bottom, but having the option to download the edit report or upload a new file can be there too.

I considered including those option under the "Verify quality edits" heading, but it didn't feel like it belonged there. I'm proposing we include it after the last edit and before the verify heading.

| Pre-verifying | Post-verifying |
| --- | --- |
| screen shot 2017-10-10 at 12 13 04 pm | screen shot 2017-10-10 at 12 13 24 pm |

The content could say: "__Download the edit report__ of the quality edits listed above. 
If the data needs to be corrected, update your file and __Upload a new file__." After the verify checkbox is selected, the second sentence can be removed.

Thoughts @keileyo ?

awolfe76:
Closes #829

- I removed the `bottom-border` on the `RefileWarning` component so that it wouldn't have the border at the bottom of the page. It still looks fine at the top too.
- The only content change made was to remove the word `below` so that the content made sense at both the top and bottom of the page.
- I also kept the warning at the bottom as an `alert` to keep it consistent.

@wpears - have a look, if you want to wait for @designlanguage to get back that's ok with me.

/cc @keileyo 

#### Syntactical and validity
syn-val

#### Quality (macro looks the same as this)
quality


wpears:
Looks good! Do we want the bottom warning for parsing errors also?

awolfe76:
I think so, give me a few and I'll add that in.

PR #876, opened 2017-11-06
Cleanup

### LINKED ISSUE #799: Clean up Institution returns
[This](https://github.com/cfpb/hmda-platform-ui/blob/f93f490133c30c6031ef2d962b032eb4d0521fe1/src/js/components/Institution.jsx#L19) and [this](https://github.com/cfpb/hmda-platform-ui/blob/f93f490133c30c6031ef2d962b032eb4d0521fe1/src/js/components/Institution.jsx#L37) have the some "container" elements. We can clean that up a bit.

awolfe76:
Closes #799

### DIFF
--- a/src/js/components/Institution.jsx
+++ b/src/js/components/Institution.jsx
@@ -14,10 +14,18 @@ const Institution = ({
   submissions,
   onDownloadClick
 }) => {
-  if (!filing)
-    return (
-      <div className="usa-grid-full">
-        <section className="institution">
+  const status = submission && submission.status
+
+  return (
+    <div className="usa-grid-full">
+      <section className="institution">
+        {/*
+        a filing should be created when an institution is created
+        so this shouldn't happen but just in case ...
+        render an alert if we don't have filings for an institution
+        otherwise show the current status
+        */}
+        {!filing ? (
           <div className="current-status">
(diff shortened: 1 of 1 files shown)


src/js/components/Institution.jsx
```diff
@@ -14,9 +14,44 @@ const Institution = ({
...
+  return (
+    <div className="usa-grid-full">
+      {/*
+        a filing should be created when an institution is created
+        so this shouldn't happen but just in case ...
+        render the current status if there is a filing
+        otherwise render an alert
+      */}
+      {filing ? (
+        <section className="institution">
```
wpears, line 28: Can we nest the check inside the section so we don't need to duplicate it?

wpears (changes requested):
One comment below

src/js/components/Institution.jsx
awolfe76: I wanted to, but [this](https://github.com/cfpb/hmda-platform-ui/pull/876/files/08c76507d1cc20a8ccfb90e000d9222b81f9791d#diff-7b62ba75715e7d849fd47779257044d6R47) being outside of the `current-status` is preventing it.

PR #882, opened 2017-11-09
Remove redux-oidc

### LINKED ISSUE #878: Remove redux-oidc
A small hand-rolled component and direct dependency on oidc-client-js would save us from bundling immutable and possibly core.js. Also, will help on the road to #792

wpears:
Instead of relying on redux-oidc, this PR removes it in favor of directly interfacing with oidc-client-js.

Needed events are listened to when the user manager is created, which dispatch appropriate actions. The user reducer has been retooled to accommodate these changes.
Closes #878


src/js/reducers/user.js
```diff
@@ -1,30 +1,53 @@
...
-  expired: false,
+  oidc: null,
+  isFetching: false,
   userError: false
 }
 
 export default (state = defaultUser, action) => {
   switch (action.type) {
-    case 'redux-oidc/SILENT_RENEW_ERROR':
+    case SESSION_TERMINATED:
```
awolfe76: Are `SESSION_TERMINATED`, `USER_SIGNED_OUT`, and `USER_EXPIRED` every handled differently?

awolfe76 (approved):

PR #891, opened 2017-11-15
Google Analytics events

### LINKED ISSUE #824: Add some event tracking to Google Tag Manager
It would be nice to track a few other things ...

- [x] when 'upload a new file' is clicked
- [x] when a new upload is triggered using the dropzone
- [x] verification (or removing verification) of quality and macro
- [x] when 'download edit report' is clicked
- [x] when 'download irs report' is clicked
- [ ] when 'history of your progress in this filing period' is clicked
- [x] when pagination is used
- [x] when signature checkbox
- [ ] submission button clicked (not sure if we need this one or not)

__Extra__
- [ ] in order to use GTM debugging tool for events we need to add `https://tagmanager.google.com` to [`nginx.tmpl`](https://github.com/cfpb/hmda-platform-ui/blob/b8e9c41ea28c258c53b3569981913b1af3dd11ae/nginx/nginx.tmpl#L26), the error now is
```
Refused to load the script 'https://tagmanager.google.com/debug' because it violates the following Content Security Policy directive: "script-src 'self' 'unsafe-inline' data: https://www.googletagmanager.com https://www.google-analytics.com".
```

@designlanguage @keileyo @wpears please add/remove from this list.

awolfe76:
Closes #824 

This adds tracking for most of what was listed in #824. 

I didn't do "in order to use GTM debugging tool for events we need to add https://tagmanager.google.com to nginx.tmpl, the error now is" because adding that led to another required security policy and because it's only necessary to get a debug view of the events ... having access to GA I could see the events as they happen. (If you'd like to see a quick demo let me know.)

I also didn't do "when 'history of your progress in this filing period' is clicked" because it isn't tied to a redux action. We can do it if we want but we can save it for another PR if we think it's necessary.

For "submission button clicked" I think we're ok not to have this because we'll easily see this from back-end metrics when we get to that point. But let me know if you think it would be good.

Thanks for the middleware suggestion @wpears. 💥


wpears:
:+1:

PR #895, opened 2017-11-16
Handle and propagate upload error

### LINKED ISSUE #890: Should bail out of upload progress when error occurs
Repro:

Upload something that will generate a 400 Bad Request
Note we get an error displayed
Note the upload bar continues when it should stop.

wpears:
Closes #890

Introduces a new action for an upload error and adds `uploadError` as a key to the upload reducer. This allows managing the validationprogress component and setting the upload state appropriately even without receiving the `UPLOADING` state from the backend (which we won't if the upload bails with, eg, a 400 Bad Request).


awolfe76:
I noticed a few things here ...

First:
- upload a bad file (`.TXT`)
- you see the error
- but the progress is still running
- if you go to the home page it says "created" (which is ok)
- going back to the upload screen, the error is gone progress is still running

Also:
- upload a bad file (`.TXT`)
- you see the error
- but the progress is still running
- upload a good file
- it works (progress continues from the point it was with the bad file), but the error stays

awolfe76 (dismissed):
See https://github.com/cfpb/hmda-platform-ui/pull/895#issuecomment-345038587

wpears:
You sure you ran the build on this branch? Your behavior sounds like how it behaves on master.

awolfe76:
Yes sir. Ignore me, forgot to run `yarn`.

wpears:
@awolfe76 Should be good to go

awolfe76 (approved):

PR #910, opened 2017-11-21
Update dockerfile and add rpms to continue using 1.10.2

wpears:
nginx 1.10.2 rpm and dependent rpms have been lifted from their build server and checked in here.

This is the same build of nginx we've used for the past several months or so, with the same flags enabled.


hkeeler (approved):
Looks good to me.  Holding off merging until official 👍 on this approach.

PR #916, opened 2017-11-22
EditsNav refactor / bugfix

### LINKED ISSUE #881: Upload Screen Misbehaving when Edits Pending
After validation has finished on a file, on a large file, there can be a gap of time between when the -ui knows that the validation has finished and when it gets the edits. During this gap of time the -ui displays that the file has syntactical and validity edits. As soon as the edits call completes it updates properly.

I saw this error when uploading this [file](https://github.com/cfpb/hmda-platform/blob/master/parser/jvm/src/test/resources/txt/clean_14117-lars.txt).

wpears:
Closes #881

Rather than having nested if statements in 3 different functions, each piece of the edits nav has been described in an object, with each entry given methods to detail whether it should display based on whether it is reachable (has the app advanced far enough to allow its display) and if so, if it has errored or completed


src/js/components/EditsNav.jsx
```diff
@@ -172,6 +19,55 @@ export default class EditsNav extends Component {
...
+        isCompleted: () => !this.props.syntacticalValidityEditsExist,
+        errorClass: 'error',
+        errorText: 'syntactical & validity edits found',
+        completedText: 'no syntactical & validity edits',
+        link: 'syntacticalvalidity'
+      },
+      'quality edits': {
+        isReachable: () =>
+          this.props.fetched &&
+          this.navMap['syntactical & validity edits'].isCompleted(),
```
awolfe76: This is causing this link to render when there are parsing errors. The [`isCompleted`](https://github.com/cfpb/hmda-platform-ui/pull/916/files#diff-07bda16239ad4cd288276a99e9a840c4R36) for `syntactical & validity edits` is `true` in this case because there are no syntactical or validity edits.

We'll probably need a `code` check as well.

parsing-nav

src/js/components/EditsNav.jsx
```diff
@@ -172,6 +19,55 @@ export default class EditsNav extends Component {
...
+        isErrored: () => !this.props.qualityVerified,
+        isCompleted: () => this.props.qualityVerified,
+        errorClass: 'warning',
+        errorText: 'quality edits found',
+        completedText: 'quality edits verified',
+        link: 'quality'
+      },
+      'macro quality edits': {
+        isReachable: () =>
```
awolfe76: Same as above.

awolfe76 (changes requested):

wpears:
Good catch, should be good to go now.

awolfe76 (approved):

awolfe76:
BTW, really like the way you changed the nav! ✨

PR #917, opened 2017-11-22
Surround editsnav with focus box

wpears:
Closes #914

PR #919, opened 2017-11-27
Use webpack import rewriting over babel

### LINKED ISSUE #792: Upgrade to React 16.0
~~### This will require a CR.~~

React recently [released 16.0](https://facebook.github.io/react/blog/2017/09/26/react-v16.0.html). This will allow us to do some refactoring because we can [return fragments and strings](https://facebook.github.io/react/blog/2017/09/26/react-v16.0.html#new-render-return-types-fragments-and-strings) but perhaps, more importantly, it will [reduce the bundle size](https://facebook.github.io/react/blog/2017/09/26/react-v16.0.html#reduced-file-size).

wpears:
As of webpack 2, it can use its own system to transpile ES6 imports, rather than leaning on babel.

This change shaves 50kb off the bundle..!

__Update__: This now closes #792
A weird cocktail of jest/babel/babel presets/webpack was causing strange errors, seemingly at random (various imports not being transpiled, dependent on jest cache, babel cache, NODE_ENV, all sorts of fun interlocking things with a bug sleeping in the middle. Bumping jest up along with react fixed this issue.


awolfe76 (approved):

awolfe76:
I'll merge after we get #916 and #917 merged.

PR #921, opened 2017-11-29
NavButton to note fetching edits

### LINKED ISSUE #915: Delay NavButton activation until edits have returned, when required
Currently, the progress blinker turns to a check as soon as validation completes, however, when edits exist, this should be delayed until the edits have returned. This will allow us to update the editsnav appropriately.

Also will kill a bug where, on a slow connection, after validation has completed but before edits have returned, using the navbutton to navigate will display improperly (show quality on the synval page, render when the main content is loading)

wpears:
Depends on #916 
Closes #915 

Leans on #916 and general routing improvements to simplify the NavButton logic why also introducing a new flag that will render a spinner when edits are needed and haven't been fetched.


src/js/components/NavButton.jsx
```diff
@@ -50,15 +44,19 @@ const NavButton = ({
...
+      key="0"
       className={`NavButton usa-button ${className || ''}`}
       tabIndex={className === 'hidden' ? -1 : 0}
       to={`${base}/${suffix}`}
     >
       {`Review ${displayName}`}
-    </Link>
-  )
+    </Link>,
+    <div key="1" className="NavSpinner">
```
awolfe76: Can we wrap this whole `<div>` in the `spinOn` check? It would just render an empty `<div>` anyway.

awolfe76 (changes requested):
See below.

awolfe76 (approved):

PR #927, opened 2017-12-01
Institutions status messages

wpears:
Companion to https://github.com/cfpb/hmda-platform/pull/1360

Closes #904, Closes #905, Closes #906, Closes #907, Closes #908, Closes #909

PR #930, opened 2017-12-06
Image updates

### LINKED ISSUE #911: Icons in banners should match the nav
Whatever icon we use in the nav bar should be repeated in the banner announcing the status.

This applies to all the banners currently. Where there's exclamation marks, they should be the same exclamation mark in a circle (used in the nav) rather than in a triangle. Once a submission is accepted, we should use the same check mark  in the banner that's used in the nav. We could use the nav color in the banner icon as well. The idea is to reinforce the steps in the nav, so we should keep the symbols consistent.

Current state:
screen shot 2017-11-22 at 11 20 09 am
screen shot 2017-11-22 at 11 24 08 am
screen shot 2017-11-22 at 11 24 35 am
screen shot 2017-11-22 at 11 25 33 am

### LINKED ISSUE #912: Change icons in nav?
Should we adjust the nav icon steps to include question marks for Quality and Macro Quality edits? Question marks seem to be more representative of what we're notifying the user of in those steps. This would have the added benefit of matching the HMDA video.

Additionally, we're currently using an X icon to represent both formatting errors and syntactical/validity edits. It could be helpful to the user to differentiate those steps further. 

| Current | Alternate |
| --- | --- |
| screen shot 2017-11-22 at 11 45 38 am | screen shot 2017-11-22 at 11 15 51 am |

### LINKED ISSUE #929: Simplify / optimize / improve handling of images
- Make sure images available from uswds aren't being needlessly duplicated
- Generate large enough png's for the check / close images that they don't seem grainy

wpears:
Depends on #927 

Closes #929 
Closes #911 
Closes #912

Uses svgs where possible and always uses the sample checks/x's/exclamations.


src/sass/components/_Alert.scss
```diff
@@ -5,19 +5,48 @@
...
+  background-size: 30px;
+}
+
+.usa-alert-error > .AlertImage {
+  background-image: url('../img/close-white.png');
+  background-image: url('../img/close-white.svg');
+  background-repeat: no-repeat;
+  background-size: 16px;
+  background-position: 7px;
+  background-color: $color-secondary;
```
awolfe76: We can remove this line to keep all the `background-color` for icons in alerts consistent.

src/sass/components/_Alert.scss
```diff
@@ -5,19 +5,48 @@
...
     width: 30px;
-    border: 2px solid $color-base;
-    background-color: $color-base;
-    background-image: url("/img/check.png");
-    background-position: center;
-    background-repeat: no-repeat;
-    background-size: 13px auto;
+    height: 30px;
+    border-radius:50%;
+    background-color: #212121;
```
awolfe76: This could be just `$color-base`.

src/sass/components/_Alert.scss
```diff
@@ -5,19 +5,48 @@
...
+  background-image: url('../img/close-white.png');
+  background-image: url('../img/close-white.svg');
+  background-repeat: no-repeat;
+  background-size: 16px;
+  background-position: 7px;
+  background-color: $color-secondary;
+}
+
+.usa-alert-warning > .AlertImage {
+  color: #fff1d2;
```
awolfe76: For these colors we should use the USWDS values, 

error = `$color-secondary-lightest`
success = `$color-green-lightest`
info = `$color-primary-alt-lightest`
warning = `$color-gold-lightest`

OR, we can set them all to `$color-white`.

src/sass/components/_EditsNav.scss
```diff
@@ -78,11 +80,12 @@
...
-          background-image: url("/img/exclamation.png");
-          background-size: 4px auto;
-          background-position: center;
-          background-repeat: no-repeat;
+          background-image: none;
           border-color: #dc731c;
+          color: #fff;
+          font-size: 30px;
+          line-height: 25px;
+          padding: 0;
```
awolfe76: Changing this to `0 1px 0 0` seems to center align it better.

There is also something different between the `!` in the nav and the `!` when its used in the alert, but I can't see what would be causing it.

awolfe76 (changes requested):

awolfe76:
@wpears - just a reminder. From conversations today, this seems like a good PR to quickly add 

1. a little more spacing between nav icons and text, and
1. the `?` for quality/macro and their alerts.

awolfe76 (approved):
I think we have it! 😄

PR #936, opened 2017-12-12
Scroll to parse errors so the user doesn't miss them

### LINKED ISSUE #913: Move formatting errors above upload box
We saw several users miss the listing of formatting errors because they didn't scroll. This is what they're seeing, and there isn't much indication that they need to look lower on the page.

screen shot 2017-11-22 at 12 03 18 pm

Even with the test data message going away when filing season opens, the list still may not make it above the fold. At this point in the process, it's important that a user understands what's wrong with their file, so I recommend showing the formatting errors above the upload box.

Uploading is still actionable above the fold with the "upload a new file" button in the error message, so I don't think we're losing anything significant by moving the box.

wpears:
Closes #913

### DIFF
--- a/__tests__/components/ParseErrors.js
+++ b/__tests__/components/ParseErrors.js
@@ -14,13 +14,17 @@ const parseJSON = JSON.parse(
 )
 
 describe('Parse errors', () => {
-  const parseErrors = ParseErrors({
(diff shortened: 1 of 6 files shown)


awolfe76 (approved):
Create #937 about the placement of the alert. But this is good to go without that update. 👍

PR #938, opened 2017-12-13
Update SubmissionPageInfo.jsx

designlanguage:
Removed "test" from the description and the period in the heading.


wpears (approved):
:+1:

PR #950, opened 2017-12-15
Youtube

### LINKED ISSUE #797: Home page video to show filing steps
_See conversation in #753 for reference._

We can put some thoughts here about what the video should say and then start the conversation with the media folks to possibly create it.

From there we can also use this issue to include the video on the homepage.

awolfe76:
Closes #797 

home-with-video


wpears (approved):
:+1:

PR #951, opened 2017-12-15
Progress

### LINKED ISSUE #570: Add progress to the institutions page
For more details see from https://github.com/cfpb/hmda-platform-ui/issues/512#issuecomment-301798018 down but the idea is to show the steps and current progress from the institutions page.

@designlanguage - I'm going to close #512 and break it out into other issues. Once we have the idea down we can re-assign this for the implementation.

awolfe76:
Closes #570 (finally!)

The text labels are hidden by default, then shown when screen hits width of 600+ ...

smaller

### DIFF
--- a/__tests__/components/InstitutionsEditsNav.js
+++ b/__tests__/components/InstitutionsEditsNav.js
@@ -1,0 +1,181 @@
+jest.unmock('../../src/js/components/InstitutionsEditsNav.jsx')
+
+import InstitutionsEditsNav from '../../src/js/components/InstitutionsEditsNav.jsx'
+import Wrapper from '../Wrapper.js'
+import React from 'react'
+import ReactDOM from 'react-dom'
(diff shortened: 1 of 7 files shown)


wpears:
Looks great!
@designlanguage will you take a look? I believe you okayed this approach but just ensuring that this instantiation is what you had in mind (specifically around us not showing the different symbols/colors.

I think the code could probably be simplified/clarified by removing the `errorClass` key altogether and just setting `navClass` directly rather than by checking that prop (the original errorClass exists because the fully-fledged edits nav needs to differentiate between warnings and errors, where we don't here).

Also, as you note in the comment about the `error` string, calling everything `error` here doesn't sound quiiiiite right, but I think would be helped quite a bit by the `errorClass` change above.

awolfe76:
@wpears - we should be good to go!

PR #969, opened 2018-01-05
No js

### LINKED ISSUE #841: No javascript
Like we do for older browsers, we could show a message if javascript is disabled.

awolfe76:
Closes #841 

If, for some reason, someone doesn't have javascript enabled this is what they will see.

I thought it would be nice to keep the idea of the header (no nav), some content (FAQs, disclaimers), and footer.

I could use some help with the content in the hero section though. For now, the heading tells them about needing javascript and I kept our intro paragraph. Any other ideas?

@wpears @designlanguage @keileyo

no-js

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -59,7 +59,7 @@ height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
             <div class="usa-width-one-whole">
               <h1>Please enable javascript to continue</h1>
               <p class="usa-font-lead">
-                Beginning with HMDA data collected in or after 2017, financial
+                In order to use the HMDA Platform, Javascript must be enabled. Beginning with HMDA data collected in or after 2017, financial
                 institutions will use the HMDA Platform to upload their
                 loan/application registers (LARs), review edits, certify the
                 accuracy and completeness of the data, and submit data for the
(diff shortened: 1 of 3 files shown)


keileyo:
I would suggest something like "Please enable javascript to use the HMDA Platform." in the header, and would add a sentence to the end of the intro paragraph restating that. 

"In order to use the HMDA Platform, javascript must be enabled."

designlanguage:
@keileyo - That heading seems a bit long. How about using @awolfe76's heading, but using your additional sentence at the beginning of the intro paragraph:

### Please enable Javascript to continue
In order to use the HMDA Platform, Javascript must be enabled. Beginning with HMDA data collected...

wpears:
I think @designlanguage 's wording makes sense. Implementation-wise I'm a little hesitant about duplicating much of the homepage's content in the noscript, but it's fine for now, especially since there won't be too much change to the homepage at this juncture.
  
Also, I added a bit of styling to fix some rendering of the analytics noscript iframe

awolfe76:
@wpears - this should be good now. See 305a7f6 for the update.

And I agree about the content. Just didn't seem too bad at the moment.

keileyo:
sounds good to me

wpears:
:+1:

PR #974, opened 2018-01-17
Request rows of edit types from Edits container

### LINKED ISSUE #971: Wait to get edits until needed
We can do something to help with https://github.com/cfpb/hmda-platform/issues/1410.

When progress finishes on an upload with call the `/edits` endpoint which returns us the list of edits found, along with their descriptions. This part is fine and gives us enough information about the edits to handle the navigation.

But, we also immediately call the `edits/<edit>` for each edit returned in the list above. This could be a lot of edits. We should wait to call for edits until the filer goes to that page. So call for syntactical and validity edits when they navigate to that page. Same for quality and macro edits ... especially in this case. We don't need quality and macro until all syntactical and validity have passed anyway.

wpears:
Closes #971 

Rather than levying this responsibility on the submissionrouter, this has been moved to the edits container to allow for finer control of edit fetching responsibilities.

A small issue re: display of EditsHeaderDescription when edit requests are coming in incrementally will be handled via https://github.com/cfpb/hmda-platform-ui/issues/975


src/js/components/EditsNav.jsx
```diff
@@ -174,5 +174,8 @@ EditsNav.propTypes = {
   syntacticalValidityEditsExist: PropTypes.bool.isRequired,
   qualityVerified: PropTypes.bool.isRequired,
   macroVerified: PropTypes.bool.isRequired,
-  fetched: PropTypes.bool.isRequired
+  editsFetched: PropTypes.bool.isRequired,
+  syntacticalValidityFetched: PropTypes.bool,
```
awolfe76, line 178: I don't see
- `syntacticalValidityFetched`,
- `qualityFetched`, or
- `macroFetched`

being used here.

src/js/components/EditsTableWrapper.jsx
```diff
@@ -102,7 +101,11 @@ EditsTableWrapper.propTypes = {
   code: PropTypes.number,
   syntacticalValidityEditsExist: PropTypes.bool,
   qualityVerified: PropTypes.bool,
-  macroVerified: PropTypes.bool
+  macroVerified: PropTypes.bool,
+  editsFetched: PropTypes.bool,
+  syntacticalValidityFetched: PropTypes.bool,
```
awolfe76, line 106: Same as above, I don't see
- `syntacticalValidityFetched`,
- `qualityFetched`, or
- `macroFetched`

being used here.

In this case I don't see
- `base`,
- `code`,
- `syntacticalValidityEditsExist`,
- `qualityVerified`, or
- `macroVerified`

either. Those are probably just lingering from an old way of doing this.

awolfe76 (changes requested):
Just a couple small things with the props, otherwise 👍 .

src/js/components/EditsNav.jsx
wpears: Those values are being provided by `SubmissionProgressHOC.jsx` to so I figured it would be better to include them in case we want to use them later and to be clear about what props are available.

src/js/components/EditsTableWrapper.jsx
wpears: Same as above, except it's `Edits.jsx` which is getting passed the props which are in turn passed down to the table wrapper.

src/js/components/EditsNav.jsx
awolfe76: Makes sense. I was just thinking about it the other way, only showing what we're using. Good with me though.

PR #976, opened 2018-01-17
Pass fetched prop to EditsHeaderDescription

### LINKED ISSUE #975: Show LoadingIcon near EditsHeaderDescription
With #974, the (X edits found) will render and be initially incorrect. This should key off the `fetched` status of a given type to either display a LoadingIcon as more results come in (preferred) or just render null until all results are in.

wpears:
Closes #975

### DIFF
--- a/src/js/components/EditsHeaderDescription.jsx
+++ b/src/js/components/EditsHeaderDescription.jsx
@@ -48,7 +48,8 @@ const EditsHeaderDescription = ({ type, count, fetched }) => {
 
 EditsHeaderDescription.propTypes = {
   type: PropTypes.string.isRequired,
-  count: PropTypes.number.isRequired
+  count: PropTypes.number.isRequired,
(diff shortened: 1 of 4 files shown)


src/js/components/EditsHeaderDescription.jsx
```diff
@@ -32,13 +33,13 @@ export const getText = editType => {
   throw new Error('Unexpected edit type. Unable to create edit description.')
 }
 
-const EditsHeaderDescription = ({ type, count }) => {
+const EditsHeaderDescription = ({ type, count, fetched }) => {
```
awolfe76, line 36: We should add `fetched` as a `propType`.

awolfe76 (changes requested):
Just the `propType`, otherwise 👍 .

awolfe76 (approved):

PR #995, opened 2018-01-24
move actions and reducers

### LINKED ISSUE #987: Move actions and reducers into new reorg
Once https://github.com/cfpb/hmda-platform-ui/pull/977 is merged, we can try to move the actions and reducers into this structure as well.

awolfe76:
Closes #987 

Give it a good look. There's a lot of reorg here.

I've been running it locally and messing around for a few hours now and it looks good. But definitely give it a run.

Basically moved actions and reducers where they seemed to fit best. There are a few directly in `./src` and even a few in `./src/utils`. The files moved to `./src` _could_ be moved into `./src/common`. Let me know what you think.

There were a few "helper" files that I moved around too, mostly into `./src/utils`, except for a few only related to testing. I moved those into `./test-resources`.


wpears:
I don't think this change fits with the overall reorg goal of reducing complexity and organizing by function.

As the state is global and actions can be fired from anywhere, I think keeping reducers in it's own folder in `src/` makes the most sense here. I don't think any one component/directory should have ownership of a reducer because such nesting doesn't have any impact on the action function of the reducers.

I'm more mixed on whether actions should be in their own folder in `src/` or not, but, if they are to be nested in component subfolders, I think they should at least be grouped together into an `actions/` directory, eg. `src/institutions/actions`. Right now component subfolders feel sprawling and random to me, and it's not immediately obvious what the purpose of all the files is.

That said, leaving `actions/` alone would be fine too, as the relationship isn't really one of ownership, even if a certain action is only called by a given component. This would save us from having to hide some actions in `utils` or `common` where they don't really fit and also allows having an exhaustive list of actions that can be dispatched in one folder (which is actually a decent benefit to keeping them together).

Thoughts?

awolfe76:
@wyatt - the short version here is that I totally agree.

I've noticed the same things jumping back and forth between this and what is currently in master now. I prefer what is in master now.

I'm fine with just closing this PR and seeing how things go for a while.

Issue #993, opened 2018-01-23
Redux improvements


awolfe76:
We implemented redux pretty early in the project. Investigate improvements that can be made to the redux implementation and state as a whole due to lessons learned.

This issue is pretty open-ended and can be used to track findings and spin-off other issues.

awolfe76:
Closing. We've done some clean up here. If we come across other things we can open new, more specific, issues.

PR #996, opened 2018-01-25
add redux devtools

awolfe76:
As part of #993 this is pretty useful.

You'll need to [install the extension](https://github.com/zalmoxisus/redux-devtools-extension#installation).

If you have another suggestion, let me know.


wpears (approved):
:+1:

PR #998, opened 2018-01-25
No edits call

### LINKED ISSUE #997: Don't call for edits unless needed
As part of #993 I noticed that we were calling for edits even though the `status` was `5: Parsed with errors`.

https://github.com/cfpb/hmda-platform-ui/blob/fc6964b8e795fa30aa6b73fcede4a7dc0d994ebd/src/actions/pollForProgress.js#L44-L46

We should prevent that, we don't need that extra call when we know there will be no edits.

awolfe76:
__Depends on #996.__

Closes #997.


wpears (approved):
Nice!

PR #999, opened 2018-01-25
WIP - Better warnings

### LINKED ISSUE #989: Move logic for edits warnings and submission receipt/info
Right now this logic is split into 2 places

The top warning or submission receipt/info is rendered here:
https://github.com/cfpb/hmda-platform-ui/blob/90967b94997e4b32dea3685d9a373833ca4a5f09/src/js/containers/Submission.jsx#L123-L132

The bottom submission receipt/info is rendered here:
https://github.com/cfpb/hmda-platform-ui/blob/90967b94997e4b32dea3685d9a373833ca4a5f09/src/js/components/Signature.jsx#L100

And the bottom edit warning is rendered here:
https://github.com/cfpb/hmda-platform-ui/blob/90967b94997e4b32dea3685d9a373833ca4a5f09/src/js/components/EditsTableWrapper.jsx#L84

Can we update this to have it all in the [`submission` container](https://github.com/cfpb/hmda-platform-ui/blob/90967b94997e4b32dea3685d9a373833ca4a5f09/src/js/containers/Submission.jsx) to make it more clear?

### LINKED ISSUE #990: Remove warning from signature component?
Do we need this anymore? They can't get here until edits have passed anyway. I suspect this has been lingering here for a while, from an old way of handling it.

https://github.com/cfpb/hmda-platform-ui/blob/f4c2c9e816d873b3d013526b6c5f868a118c194c/src/submission/signature/index.jsx#L16-L35

awolfe76:
### WIP

__Depends on #996 and #998__

Closes #989 
Closes #990 

For the edit warnings, I moved the rendering into the `tableWrapper`. They wouldn't work all in the submission `container` because of the rendering of the verifier, which also happens in the `tableWrapper`.

For the `pageInfo` and `receipt` on the submission page, I moved those all into the submission `container` because there are multiple components rendered on that page. 

I've also removed some of the logic where we checked for `status` in the submission `container`. We shouldn't need to do that because those pages can't be accessed unless that status was true already. I imagine it was another piece that was just lingering from the old way of doing it.

### DIFF
--- a/src/submission/container.jsx
+++ b/src/submission/container.jsx
@@ -6,11 +6,11 @@ import fetchSubmission from '../actions/fetchSubmission.js'
 import fetchInstitution from '../actions/fetchInstitution.js'
 import setFilename from '../actions/setFilename.js'
 import UserHeading from './UserHeading.jsx'
-import SubmissionPageInfo from './ReadyToSign.jsx'
+import ReadyToSign from './ReadyToSign.jsx'
 import UploadForm from './upload/container.jsx'
 import ErrorWarning from '../common/ErrorWarning.jsx'
 import EditsContainer from './edits/container.jsx'
-import SubmissionReceipt from './ReceiptContainer.jsx'
+import ReceiptContainer from './ReceiptContainer.jsx'
 import EditsNavComponent from './Nav.jsx'
 import NavButtonComponent from './NavButton.jsx'
 import RefileWarningContainer from '../refileWarning/container.jsx'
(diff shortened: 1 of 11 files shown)


awolfe76:
Should have mentioned, the reason I started down this path was because I noticed 2 calls being made to `/sign`. This no longer happens.

Added WIP, because now it does __3 calls__. 😞

src/submission/container.jsx
```diff
@@ -47,15 +47,23 @@ const renderByCode = (code, page, message) => {
...
+      toRender.push(<Edits />)
     } else if (page === 'submission') {
-      if (code > VALIDATING) {
-        toRender.push(<IRSReport />)
-        toRender.push(<Summary />)
-        toRender.push(<Signature />)
+      let warningOrReceipt = null
+
+      if (code !== SIGNED) {
+        warningOrReceipt = <ReadyToSign />
```
wpears: I think I prefer having the ReceiptContainer below the Signature component so it matched how the quality and macro verifiers work (and in general I think it's more logical for a new thing to be inserted below/at the bottom of the page). I missed the receipt and thought the request was hanging on my first submission.

awolfe76: Yep, just wasn't thinking about that part. Will fix that.

PR #1002, opened 2018-01-26
Add verified check in router + tests and bonus nav tests

### LINKED ISSUE #1001: Macro Edits page - wrong refresh behavior?
If I'm on the macro edits page (not verified) and I refresh the page, it goes to the quality edits page. 

(Maybe real users don't refresh the page all the time, so they wouldn't run into this? I do, since I'm testing out back-end API calls)

wpears:
Closes #1001

### DIFF
--- a/src/submission/Nav.test.js
+++ b/src/submission/Nav.test.js
@@ -241,3 +241,139 @@ describe('render', () => {
     )
   })
 })
+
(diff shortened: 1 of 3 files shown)


awolfe76 (approved):
👍

PR #1004, opened 2018-01-29
Prevent rerenders

### LINKED ISSUE #1003: Use why render to remove unnecessary renders
Especially when uploading a new file to remove jank during the animation. Also seems like submissions page has some rerenders that can be avoided.

wpears:
Closes #1003


src/utils/shouldComponentUpdate.js
```diff
@@ -0,0 +1,16 @@
+import isEqual from 'lodash.isequal'
+
+export default function shouldComponentUpdate(nextProps) {
+  const props = this.props
```
awolfe76: I think a comment here about how `this.props` is available would be nice.

awolfe76:
This all looks good!

I did notice another place where this happens, during the initial load of the institutions page (and if you hit refresh) you'll see the `Institutions` and `InstitutionsContainer` are re-rendered.

_Actions added just to see the steps._
```
action @ 09:06:09.701 RECEIVE_FILING
Institutions
  Institutions.props: Value did not change. Avoidable re-render!
  Institutions.state: Value is the same (equal by reference). Avoidable re-render!
InstitutionContainer
  InstitutionContainer.props: Value did not change. Avoidable re-render!
  InstitutionContainer.state: Value is the same (equal by reference). Avoidable re-render!
action @ 09:06:09.742 RECEIVE_FILINGS
```

Want to take a look at that one here too? Or open a new issue?

__EDIT__: this happens when a filer has multiple institutions.

awolfe76 (approved):
👍

Issue #1028, opened 2018-02-06
Cleanup and improve readability


awolfe76:
There are a few places where the logic is confusing. It would be good to read through and make sure these can be easily understood (as best we can).

For example:
https://github.com/cfpb/hmda-platform-ui/blob/82174c7888d35eb650d306e4babf4e770712709c/src/institutions/index.jsx#L44-L76

awolfe76:
Closing. We've done some clean up here. If we come across other things we can open new, more specific, issues.

PR #1029, opened 2018-02-06
Refactor

awolfe76:
Part of #1028 ... more to come.

This tries to make a few things more readable.

But it also does a good bit of work on the `institutions/index` and `ValidationProgress` to clean them up. There are a few more components that I'm going to move to next but I'll put those in a separate PR to keep things smaller.


wpears (approved):
Cool. Easier on the eyes all around

PR #1030, opened 2018-02-06
add active style for dropzone to show that it's ready

### LINKED ISSUE #1021: Highlight upload area when hovered with a file
We can do this before v2.

When someone hovers over the drop area we should highlight the area, eg background color change, border change, etc.

awolfe76:
Closes #1021 

dropzone

### DIFF
--- a/src/submission/upload/_UploadForm.scss
+++ b/src/submission/upload/_UploadForm.scss
@@ -87,4 +87,9 @@
       }
     }
   }
+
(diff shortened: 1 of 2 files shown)


wpears:
Any thoughts on making the color change on a plain hover as well?

awolfe76:
I think that's a good idea, it will definitely help people who use the click vs the drag-n-drop.
Added in 52a6341.

PR #1034, opened 2018-02-06
Data warning

### LINKED ISSUE #1009: Add a warning for Loan or Application Numbers regarding privacy
The HMDA regulations strongly encourage filers to avoid using customer personal information in the loadID field.  In order to remind filers and highlight potential concerns, we need to make two changes to the platform:

1. Post a notice to filers, on the platform or institution landing page (or similar). This can be done ahead of the next item. 

2. The backend will be able to detect loadIDs that are all alpha characters or ###-##-#### pattern when [#1464](https://github.com/cfpb/hmda-platform/issues/1464) ticket is complete. This will work in the same way as a quality edit. To support this work, we need to design how the alert will appear, the language in the edit, and the user interaction (same as quality edits, unique, etc.). 

For both, the guidance comes from [1003.4(a)(1) comment 4](https://www.consumerfinance.gov/eregulations/1003-4/2015-26607_20170101#1003-4-a-2) which states in regards to Application or Loan Number:  
>Institutions are strongly encouraged not to use the applicant's or borrower's name or social security number, for privacy reasons.

awolfe76:
Closes #1009 

I didn't want to use an `Alert` because I didn't want it taking up too much space above the input. I don't think it has to dominate the page.

data-warning


src/institutions/Header.jsx
```diff
@@ -18,10 +18,6 @@ const InstitutionsHeader = ({ filingPeriod }) => {
         <p className="usa-font-lead">
           You may file HMDA data for your authorized institutions below.
         </p>
-        <p>
-          Your progress will be saved if you leave the platform before
```
wpears, line 23: Is there some guidance on removing this language?

awolfe76: Sorry. I should have created an issue for this a while ago.

That language isn't useful in this location. It's much more useful during the upload process, where we also have it for large files. Here it just adds more noise.

I had talked to @designlanguage a while back about it and then forgot to create the issue, but he agreed.

Then, in other more recent conversations with @jmarin, it came up that it might not be a good idea to say this up front because we __need__ filers to allow at least the upload to finish. When we show it during the upload process the actual upload is already complete.

awolfe76:
@wpears - hold of on merging this one. The PR on the back-end for this edit was merged so I'm going to finish this up with that work.

awolfe76:
_Removing WIP_

## Initial Warning

I've updated the warning on the upload, just a slight color change. I thought it should match the highlight for Q666 (more in a minute). 

upload-warning

## Q666

I've highlighted Q666, just using a simple background color (warning) and some padding. But, I've also updated the `<caption>` and `.caption` a bit to improve the spacing. (I think there's more to do here but this seems like a good first step.)

I've left Q666 at the bottom of the page, rather than moving it to the top. I think it's actually more noticeable closer to the verification checkbox. But, I'm open to suggestions on styling to highlight it. I'm not sure if using the same color as the "your file has edits" warning is a good idea ... maybe it's fine. Thoughts? @designlanguage @wpears

### Full screenshot showing updated captions and highlighted Q666

quality-edits

wpears:
I like the changes, I'd be interested to hear George's thoughts on the same-color warning though.

designlanguage:
@wpears - The same coloring is tricky. We've established red to mean that something is wrong and it can't proceed until fixed. This seems more like a yellow alert.

The downside to having the same color is that it could seem like repeated information. If this edit being highlighted because it needs more attention than the others, can it be included in the first alert on the page? As is, that would be a lot of content, but would it help for a filer to see it right away?

awolfe76:
@wpears - b5ab7a9 should finish this one off.

@designlanguage - I think based on the conversation earlier you are ok with this as is, right?

PR #1035, opened 2018-02-06
update verified state based on API response

### LINKED ISSUE #1014: Edit verification - use the response instead of the checked value
When edits are verified we're setting the state based on the value of the checkbox. 

https://github.com/cfpb/hmda-platform-ui/blob/7e45594d100ab69fee2a825ad5b878c2f5eaf92a/src/actions/fetchVerify.js#L25-L26

We should instead use the value in the JSON response from the API, which looks like:

```json
{
  "verified": true
}
```

This seems like it would be keeping us better in-line with the back-end.

awolfe76:
Closes #1014 

02f66c7 just updates to use `propTypes`. I started seeing this warning when running jest ...
`Warning: Component InstitutionRefile declared PropTypes instead of propTypes. Did you misspell the property assignment?`

Sorry, added 6890069 to remove a weird `<div>` from the home component. Not sure why we had that.

### DIFF
--- a/src/institutions/Institution.jsx
+++ b/src/institutions/Institution.jsx
@@ -71,11 +71,11 @@ const Institution = ({
   )
 }
 
-Institution.PropTypes = {
+Institution.propTypes = {
   institution: PropTypes.object,
   filing: PropTypes.object,
   submission: PropTypes.object,
-  submissions: PropTypes.object,
+  submissions: PropTypes.array,
   onDownloadClick: PropTypes.func
 }
 
--- a/src/actions/fetchVerify.js
+++ b/src/actions/fetchVerify.js
@@ -22,8 +22,8 @@ export default function fetchVerify(type, checked) {
             throw new Error(json && `${json.status}: ${json.statusText}`)
           }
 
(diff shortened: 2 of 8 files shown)


src/institutions/Institution.jsx
```diff
@@ -71,11 +71,11 @@ const Institution = ({
   )
 }
 
-Institution.PropTypes = {
+Institution.propTypes = {
   institution: PropTypes.object,
   filing: PropTypes.object,
   submission: PropTypes.object,
-  submissions: PropTypes.object,
+  submissions: PropTypes.array,
```
awolfe76, line 78: Not sure why this wasn't triggered before, but React started to warn that we were passing an array (which we are) instead of an object. Maybe just didn't notice it before.

And we had it correctly here (the only placed is passed down too):
https://github.com/cfpb/hmda-platform-ui/blob/82174c7888d35eb650d306e4babf4e770712709c/src/institutions/SubmissionHistory.jsx#L87

wpears: Wasn't warning before because of the incorrect capitalization ^^; Good catch

PR #1039, opened 2018-02-08
Scaffolding for api refactor

wpears:
Ensures id, filing, and submission are available from the submission router, so API calls can be made simpler and less state needs to be passed around (to components, actions, etc) and pulled from the URL.

In another PR, I'll be simplifying the api function signatures, removing unnecessary props, and removing references to `location`.


src/api/fetch.js
```diff
@@ -1,19 +1,32 @@
 import isomorphicFetch from 'isomorphic-fetch'
 import createQueryString from './createQueryString.js'
-import parseLocation from './parseLocation.js'
 import makeUrl from './makeUrl.js'
 import * as AccessToken from './AccessToken.js'
 import { signinRedirect } from '../utils/redirect.js'
 import log, { error } from '../utils/log.js'
 
+let store = null
```
awolfe76: I think a comment here about how this is called from `index.js`, after the store is created there, would be helpful.

src/api/api.js
```diff
@@ -19,12 +19,8 @@ export function getFiling(id, filing) {
   return fetch({ pathname: `/institutions/${id}/filings/${filing}` })
 }
 
-export function getFilingFromUrl() {
-  return fetch()
-}
-
 export function getLatestSubmission() {
-  return fetch({ suffix: '/submissions/latest' })
+  return fetch({ submission: 'latest' })
```
awolfe76, line 23: This one took me a minute to see what was happening in `fetch` but 👍 .

awolfe76 (changes requested):
This looks good! Just one small comment about adding a ... comment in `fetch.js` for the `store`.

Looking forward to seeing what you do with the api functions.

awolfe76 (approved):

PR #1041, opened 2018-02-09
Refactor api signatures

wpears:
Depends on #1039
Removes submission keys from api and also removes submission setter/getter which was secretly nestled in the actions


awolfe76 (approved):

awolfe76:
Awesome!

Issue #840, opened 2017-10-18
On upload progress show the %


awolfe76:
Rather than showing the `0%` and `100%`, which I think are understood, we can show the actual % in a small "popover".

PR #1043, opened 2018-02-09
Upload

### LINKED ISSUE #1040: Show warning about closing window during upload
After the upload step of a submission, we tell the filer that they can close the window and come back later.

We should add some content __during__ the upload step that says something like "please do not close your window until the upload is complete" and then replace it with the content above.

We've seen cases where this has happened with large files and it would be good to be clear about what is required.

awolfe76:
Closes #1040.

Also, to work toward #840 I refactored the `/upload` to improve readability. I split out several pieces into new components, overall making them much more manageable.


wpears:
Cool, changes look good, always a fan of simpler modules :)

PR #1044, opened 2018-02-13
Render an error on oidc-call errback

wpears:
repro to see an error:
visit <app-url>/oidc-callback#state=123

This will allow users with strange and unexpected errors (like local computer time causing failing token validation) to report a more complete picture to HMDA Help or fix it themselves.


src/oidcCallback.jsx
```diff
@@ -12,7 +14,31 @@ export class oidcCallback extends React.Component {
   }
 
   errorCallback(e) {
-    browserHistory.replace('/')
+    this.props.dispatch(receiveError(e))
```
awolfe76: Wouldn't we need `mapDispatchToProps` for this to work?

awolfe76 (changes requested):
Other than the one comment, where is [`signinRedirectCallback()`](https://github.com/cfpb/hmda-platform-ui/pull/1044/files#diff-98d7383c2a98e80ba17627b0dea13f19R50) coming from (even though it wasn't part of this PR)?

src/oidcCallback.jsx
```diff
@@ -12,7 +14,31 @@ export class oidcCallback extends React.Component {
...
+  }
+
+  renderError() {
+    return (
+      <Alert type="error" heading={this.props.error.message}>
+        <div>
```
awolfe76: I think this is a good place to use `<React.Fragment>`.

src/oidcCallback.jsx
wpears: Actually, in a connected component, if `mapDispatchToProps` isn't provided, it will similar pass through the dispatch function as a prop. I'd been meaning to clean up some places where we define and pass the function, but it only returns `dispatch`, so I tacked that on to this PR :)

src/oidcCallback.jsx
wpears: I actually did started to do that but my editor didn't like `<>` so I abandoned it out of protest ;)
Also, I wasn't sure it would work due to how we're rendering children inside the Alert, but it actually works fine ^^;

wpears:
`signinRedirectCallback` is a method on the user manager, which is part of `oidc-client-js`. It gets created in `utils/createUserManager`.

awolfe76 (approved):

PR #1048, opened 2018-02-16
Suppress tables

### LINKED ISSUE #1018: Logic difference between navButton and progress navigation
From https://github.com/cfpb/hmda-platform/issues/1477.

Just wanted to make sure to get this into an issue so we can think about different ways of handling this in the UI. _The current way may be fine._

I think we have some mismatched logic between how the nav gets rendered and how the "next" buttons at the bottom get rendered.

When this weird state from https://github.com/cfpb/hmda-platform/issues/1477 happens you'll see that the "submission" link in the nav is not available, which is good given the status, but, they can still get there by using the "next" buttons. It seems like that button shouldn't be available either and we should render an error instead because we have a state that doesn't match the edit count.

But ... maybe what is happening is ok? There's an error message and the signature is disabled so maybe we're ok?

### LINKED ISSUE #1046: Don't render edit details for large files
When we have a large file uploaded

1. make the initial call to `/edits` but
1. don't make calls to specific edits, `/edits/<edit>`, instead
1. make sure the edit pages render high-level details for the edits found (from `/edits`), AND
1. highlight the `CSV` download instead.

wpears:
Closes #1046 
Closes #1018


awolfe76 (approved):
👍

PR #1052, opened 2018-02-28
Retry on polling

### LINKED ISSUE #1047: Make additional calls when polling
Sometimes, during a file upload and processing, a polling call with fail and the UI will display an error message and stop polling. This is fine but, the solution to this problem is often to have the user refresh the page (we even say this in the UI error). 

Instead of failing immediately it might be better to continue to poll, at least a few more times. If a refresh can fix the problem, making the API call again should do the trick too.

_During the conversation about this @hkeeler said he noticed that at times the calls were being made without a token. For this case, we can take a similar approach as above but if we see a `503` we could refresh the token first, or show the "logged out" modal, and then make another API call._

wpears:
Closes #1047 

Also fixes a subtle bug where an old poll would be processed once. This was non-destructive, but unnecessary.


awolfe76 (approved):
👍

PR #1056, opened 2018-03-09
Increase coverage

wpears:
Stopping for now due to an annoying bug related to https://github.com/istanbuljs/istanbuljs/issues/119

Once a few projects downstream of that update, we can bump up several of our devDeps and function coverage should jump up a bit.


awolfe76 (approved):
👍

PR #1057, opened 2018-03-15
Select filing period

### LINKED ISSUE #1016: Handling year changes
_This is specifically for V2 and would need to be merged into the `v2.0.x` branch._

Things to consider:
- [ ] how this is handled in the UI
- [ ] code changes required to make this work

Starting next filing season filers will need to be able to view, upload for the first time, re-submit, etc. They will have the same functionality no matter what year they select. 

_Keep in mind, filers can be asked to resubmit 2 years back._

### UI

By default, we should have them enter the app in the current filing period.

We need to give them the ability to switch to the previous year. In my head right now, I see this as a simple drop-down option somewhere in the UI. I think some prototypes here would be great. This doesn't have to be complex.

We also need to make it very clear within the UI which year they are currently viewing.

Does this require special messaging in the confirmation modal if we notice a new upload for a year that isn't the current year?

Should the back-end return a special message if we see a file uploaded to `/2017` but the data looks like it's for 2018, eg "it looks like you are trying to file a 2018 submission for 2017"?

### Code

I don't think we're in bad shape here, but we'll have to dig a bit to be sure.

We're already sending requests with the filing period included, `/institutions/0/filings/2017/submissions/3/edits`. So as long as we update the filing period state when someone switches years we _might_ be ok. Again, more digging to do.

Another thing to think about is that we might want to keep state by year, but I don't think this is necessary ... making new requests if they switch years isn't bad here because I don't think it's that likely to happen. And even if it does happen a lot it's still not bad and helps keep the state saner.

@wpears @designlanguage @keileyo @colleenms

### LINKED ISSUE #1055: The filing period is closed messaging
Some users are interpreting the following message to mean they can begin testing 2018 LAR data in the platform: 

**The filing period is closed.**
The HMDA Platform remains available outside of the filing period to upload, test, and validate HMDA data.

Suggest updating this to the following until the beta release of the 2018 Platform:

**The filing period is closed.**
The HMDA Platform remains available outside of the filing period for late submissions and re-submissions of 2017 HMDA data.

@keileyo

wpears:
Closes #1016

There are some subtle issues we need to talk through, but this gets us in a good spot to do so:
  - Will filings be initialized for past filing periods for new filers? If not what should we display?
  - When does the default switch over to 2018? How do we want to message the beta period of 2018 when there is a live version of 2017?
  - Before the 2018 beta period, what/how should we message for 2018 filing if someone switches to 2018 out of curiosity?

An attempt was made in this PR to answer each of these questions is a reasonable way, but it'll be good to chew the fat a bit on this.

### DIFF
--- a/src/actions/fetchCurrentFiling.js
+++ b/src/actions/fetchCurrentFiling.js
@@ -6,6 +6,6 @@ export default function fetchCurrentFiling(filings) {
     const filing = filings.filter(filing => {
       return filing.period === period
     })[0]
-    return dispatch(fetchFiling(filing))
+    if (filing) return dispatch(fetchFiling(filing))
   }
 }
--- a/src/actions/fetchCurrentFiling.test.js
+++ b/src/actions/fetchCurrentFiling.test.js
@@ -25,7 +25,7 @@ describe('fetchCurrentFiling', () => {
     store.dispatch(fetchCurrentFiling(filings)).then(() => {
       setTimeout(() => {
         expect(store.getActions()).toEqual([
-          { type: 'REQUEST_FILING', id: '123' },
+          { type: 'REQUEST_FILING', id: '123', period: '2017' },
           { type: 'RECEIVE_FILING', filing: { filing: 'afiling' } }
         ])
         done()
--- a/src/actions/fetchEachInstitution.test.js
+++ b/src/actions/fetchEachInstitution.test.js
@@ -23,7 +23,46 @@ fetchCurrentFiling.mockImplementation(() => () => {
 })
 
 const mockStore = configureMockStore([thunk])
+const filings = [
+  {
+    institutionId: '0',
+    period: '2017',
+    status: { code: 1, message: 'not-started' }
+  },
+  {
+    institutionId: '0',
+    period: '2016',
+    status: { code: 3, message: 'completed' }
+  }
+]
 
+const f2 = [
+  {
+    institutionId: '1',
+    period: '2017',
+    status: { code: 3, message: 'completed' }
+  }
+]
+
+const f3 = [
+  {
+    institutionId: '2',
+    period: '2016',
+    status: { code: 3, message: 'completed' }
+  }
+]
+const f4 = [
+  {
+    institutionId: '3',
+    period: '2017',
+    status: { code: 1, message: 'not-started' }
+  },
+  {
+    institutionId: '3',
+    period: '2016',
+    status: { code: 3, message: 'completed' }
+  }
+]
 const getEachInstitution = [
   { type: types.REQUEST_INSTITUTION, id: '0' },
   { type: types.REQUEST_INSTITUTION, id: '1' },
@@ -31,19 +70,23 @@ const getEachInstitution = [
   { type: types.REQUEST_INSTITUTION, id: '3' },
   {
     type: types.RECEIVE_INSTITUTION,
-    institution: institutionsDetailObj['0'].institution
+    institution: institutionsDetailObj['0'].institution,
+    filings: filings
   },
   {
     type: types.RECEIVE_INSTITUTION,
-    institution: institutionsDetailObj['1'].institution
+    institution: institutionsDetailObj['1'].institution,
+    filings: f2
   },
   {
     type: types.RECEIVE_INSTITUTION,
-    institution: institutionsDetailObj['2'].institution
+    institution: institutionsDetailObj['2'].institution,
+    filings: f3
   },
   {
     type: types.RECEIVE_INSTITUTION,
-    institution: institutionsDetailObj['3'].institution
+    institution: institutionsDetailObj['3'].institution,
+    filings: f4
   }
 ]
 describe('fetchEachInstitution', () => {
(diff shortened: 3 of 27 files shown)


awolfe76:
I've started to review this and so far, so good. But, can we break this out into 2 PRs, one to handle #1055 and another to handle #1016?

I think it would be good to get #1055 out ASAP (this release) but #1016, like you mention, will take a bit more time to figure out.

#1016 would probably be best as a PR to the `v2` branch too. Although I can see a reason to do against `master` as a "coming soon" announcement. Something else to think through.

awolfe76:
> Will filings be initialized for past filing periods for new filers? If not what should we display?

That's a good question. I think our current message `No filing initialized for <year>. Your filing has not been initialized for filing period <year>. If you need to submit data for this filing period, please contact HMDA Help` does the trick. Maybe we could get a little more specific but I think that's ok for now.

> When does the default switch over to 2018? How do we want to message the beta period of 2018 when there is a live version of 2017?

This one is interesting. I'm going to take a guess here and say that we should do the default switch as soon as the beta is ready. I think that we'll see more traffic to beta than we will for 2017, almost immediately, which would justify the default change. (Or, we could wait until that's proven in analytics.)

I understand why the header navigation is a good option for placement of the year choices, but this is also why part of the reason why I'm not sure it's the best option. There are a couple of things here:

- we'll only ever have 3 years available; current and 2 years prior
- 412 filers refiled data in 2016, ~5% (while more than I thought it might be it's still pretty low).

All of that makes me think it might not be a bad idea to have the options on the institution listing page, after login. 

#### _Quick example (content needed of course)_
year-options

Or, maybe better yet would be to have the "2018 beta" in that screenshot actually be the drop-down. We did this on the [rural and underserved tool](https://www.consumerfinance.gov/rural-or-underserved-tool/). That seems like it would be easy to use and very visible/clear. (And yeah, the option would only show up once logged in. If we keep the options in the header I think that should apply there as well.)

Also, during an actual filing, I think it's very unlikely someone would want to switch periods. So having the option available on the logged in home/landing page makes sense for this.

> Before the 2018 beta period, what/how should we message for 2018 filing if someone switches to 2018 out of curiosity?

I don't think we should push out this option until we're totally ready for the beta period unless we think it would be valuable as a "coming soon". But I would vote to wait.

colleenms:
Will filings be initialized for past filing periods for new filers? If not what should we display?

Yes, but the number of filers will likely be very small. @awolfe76 will the message you referenced be used only once the default is switched to 2018? this may also be impacted by whether user accounts in keycloak persist or will be deleted at some point. if the former is true, at what point will an individual be considered a "new filer" for 2018?

When does the default switch over to 2018? How do we want to message the beta period of 2018 when there is a live version of 2017?

Agree with @awolfe76 ; this should happen when beta is released. We should also think about how this is different for a filer with multiple institutions. Is the year selection applied to the individual institution or the full list of institutions?

Before the 2018 beta period, what/how should we message for 2018 filing if someone switches to 2018 out of curiosity?

Agree with @awolfe76 ; don't think we should show this until the user can actually upload 2018 data

designlanguage:
I definitely think we need the year selectable on the institutions page.

I'm interested to know the benefits of having a selectable year in the header. If someone is in the process of filing, what would happen if they then selected another year? Would it kick them to the institutions page for that year? If so, that seems a bit jarring.

@awolfe76 - I like the idea of something like the selector in the [Rural and underserved area tool](https://www.consumerfinance.gov/rural-or-underserved-tool/). I also think we need visual differentiation for closed filing periods, open periods, and beta.

Here's a sketch showing how that might play out:

| Previous year | Beta period |
| --- | --- |
| screen shot 2018-03-20 at 2 49 41 pm | screen shot 2018-03-20 at 2 49 31 pm |
| Previous years could have all text in dark gray. The institution box could be outlined, and the year selector could have a light gray background. | During beta, the year selector could have a background color that matches the banner, and even say "beta" next to the year. |

awolfe76:
@colleenms 

> will the message you referenced be used only once the default is switched to 2018?

That message would apply to anyone who didn't have a filing created. It does this today too. This is very rare. When we create institutions in the system as a filer, they get a filing created for that year. We'll just have to be sure this is done for 2018 like it was for 2017.

---

@designlanguage - When the 2018 filing period opens, the UI is basically what it was for the 2017 filing period (no yellow, no dark gray) and the 2017 filing period stays dark gray?

designlanguage:
@colleenms - Correct, something like this:

screen shot 2018-03-20 at 4 50 59 pm

wpears:
Do we have consensus on the action items here? I can remove the 2018 option from the selector (though maybe we should keep it for now if this is just going on the v2 branch?)

Should we go with https://github.com/cfpb/hmda-platform-ui/pull/1057#issuecomment-374752238 over the selector in the header? I agree that it's unlikely people will want to switch mid-filing and will typically end up being bumped to the institutions page anyhow (or at least to `/upload`) so it makes sense to put it there.

awolfe76:
Yeah, the consensus seems to be using the screens from https://github.com/cfpb/hmda-platform-ui/pull/1057#issuecomment-374719921 and https://github.com/cfpb/hmda-platform-ui/pull/1057#issuecomment-374752238 ... putting the options on the institutions page, putting the drop-down in the heading.

And I'd say leave the 2018 option and we'll merge into v2. We'll just have to start maintaining that branch as long as we need to.

awolfe76:
@wpears - should we also update the [`filingPeriod` reducer](https://github.com/cfpb/hmda-platform-ui/blob/5c6b19eff9502c8b7c8fb7f8385e288ef7dff628/src/reducers/filingPeriod.js#L5) to default to `2018`?

Issue #1025, opened 2018-02-02
Improve IA around all of HMDA content and tools


awolfe76:
We can move most of, if not all, content to ffiec.cfpb.gov ... along with our tools currently running on gh-pages (including API docs).

This will improve the overall UX for people doing anything with HMDA.

A quick thought (in no particular order):

- /
- /tools
  - /check-digit
  - /rate-spread
  - /file-format-verification
- /reports
  - /aggregate
  - /disclosure
  - /modified-lar
- /for-filers
  - /filing-instruction-guides
  - /lar-formatting-tool
  - /faq
  - /tech-preview
- /developers
- /blog

Does the app path need to change, /filing?

designlanguage:
@awolfe76 - Most of this content is split up into the type of content, but "for filers" is split by audience. That makes the LAR formatting tool an outlier. I'd guess that many people would expect the LAR formatting tool to be under tools. To avoid confusion, how about we keep it split by content type? That would mean reframing "for filers" as "help" or "instructions". Even "instructions for filers" would be fine.

Also, does the current platform login (homepage) need to be lumped in with all this material or can it be separate like each of these other sections? I prefer separate, linked on this page, and maybe it becomes a banner at the top of the page during filing season. Interested to know your thoughts on this:

- /
- /platform
- /tools __- could be framed as supporting tools__
  - /check-digit
  - /rate-spread
  - /file-format-verification
  - /lar-formatting-tool
- /reports
  - /aggregate
  - /disclosure
  - /modified-lar
- /for-filers __-could be reframed as help or instructions for filers__
  - /filing-instruction-guides
  - /faq __- consolidate with knowledgebase?__
  - /tech-preview
- /developers
- /blog

designlanguage:
I don't think the top nav of the platform should be conflated with the nav of the homepage. Using the USDS web standards for menus, that might look something like this:

![screen shot 2018-03-09 at 5 44 53 pm](https://user-images.githubusercontent.com/10562544/37233699-41cc8ef2-23c2-11e8-9688-bc34028d3ab5.png)

Three of the menus could have drop-downs (for where there are subsections). For example, that would allow you to navigate to a specific tool from the homepage.

PR #1059, opened 2018-03-16
Nav updates

awolfe76:
#### The has several PRs that go with it and should be merged AFTER others.

---

Adds in links to tools and data publication for logged out users.

Adjusts logged in navigation by moving "Home" to the left of the user ...

nav

### DIFF
--- a/src/common/Header.jsx
+++ b/src/common/Header.jsx
@@ -18,18 +18,17 @@ export const logOutHandler = e => {
 export const makeNav = (props, page) => {
   let userHeader = (
     <ul className="usa-nav-primary">
-      {props.user ? <li>{props.user.profile.name}</li> : null}
+      <li>
+        <HomeLink />
+      </li>
       {props.user ? (
-        <li>
+        <li className="user">
+          {props.user.profile.name}
           <a href="#" className="usa-nav-link" onClick={logOutHandler}>
             Logout
           </a>
         </li>
-      ) : null}
-      <li>
-        <HomeLink />
-      </li>
-      {props.user ? null : (
+      ) : (
         <React.Fragment>
           <li>
             <a href="/data-publication/" className="usa-nav-link">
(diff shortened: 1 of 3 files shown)


wpears:
Seems to me like we'd want to always display the data-publication and tools links in the header. I don't think it's intuitive that you'd need to log out in order to get those nav options.

wpears (changes requested):
*(see above)*

awolfe76:
I definitely see what you mean here.

The goal was to make the platform, the app, be outside of the rest of the site. We can, and need to, do a little more to make sure this separation is a bit more clear. Things like:

- a little bit different styling,
- maybe make the app link open a new window, and
- possibly a few others.

There are other updates coming to the home page to call out other parts of the site, as well as the platform.

@designlanguage @keileyo @colleenms

designlanguage:
@wpears - There will be a lot of content on the home page accessible to the public (non-filers) that doesn't have anything to do with logging into the platform, so having the tools/resources as pages within the platform nav doesn't seem accurate.

It gets tricky with what the header looks like as well. Would the main text in the header always say "FFIEC HMDA Platform"? If so, that could be confusing for non-filers.

To help the user understand where they are, I recommend differentiating the homepage header from the platform. I also don't think a filer would need to log out to access other resources. When they select the platform from the homepage, it could open a new tab/window into the app. The platform is complex enough with the step navigation, and it'll only get more complex with multiple years being introduced, so I think we should keep it simple and keep the app separate from the homepage.

Here's what I'm recommending: https://github.com/cfpb/hmda-platform-ui/issues/1025

If there are benefits I'm not considering, let's take time today to discuss.

wpears (approved):
Approved as a stopgap for 4/1 before https://github.com/cfpb/hmda-platform-ui/issues/1025#issuecomment-371968362 :+1:

PR #1061, opened 2018-04-03
New loading icon

### LINKED ISSUE #1054: Switch to pure CSS spinner
Along the lines of https://atomiks.github.io/30-seconds-of-css/#donut-spinner

wpears:
Depends on #1057 
Closes #1054
Switches to a pure CSS loading icon, classing up the loading icon a bit and saving on bytes (you shouldn't have to load something to see that you're loading something).

With https://github.com/cfpb/hmda-platform-tools/pull/129 and https://github.com/cfpb/hmda-platform-auth/pull/226


awolfe76:
@wpears - The icon looks 🌟 . I did notice a couple things while testing it out though, mostly unrelated to this PR but wanted to see what you thought about adding to it or creating new issues.

1. On the ~~home~~ institutions page, if you slow down the network, the loading icon position is ... odd. :)
homepage-loading
1. When loading the IRS, it seems like we'd be better to have the IRS heading and description showing and have the loading icon where the table will render.
1. I think we also need a loading icon on submission page with the "Submit HMDA data" button. I noticed this after slowing down the network and submitting a file. It seemed like nothing was happening.

Each of those could be separate issues, let me know.

---

Also, do you want to get this in now? There's nothing preventing rolling this one out soon. Should we remove the dependency on #1057? #1057 could take a bit longer and it should go against v2 anyway. I'm ok either way on this one too, just wanted to point out that it could be a little longer if we wait.

wpears:
Issues 1, 2, and 3 should be fixed (along with some other secret ones).

I'm okay waiting on this, though to my mind, we'd want #1057 to go to master so we can start messaging the closed filing period/select filing period stuff. (ie, if someone visits the app and tries to  filing for the current year, they'll be able to select 2018 and immediately see they can't file yet).

PR #1069, opened 2018-05-11
Move ui to /filing/

### LINKED ISSUE #1071: Header updates
Remove general header and use one specific to the filing app (this may be the same module with different children/props passed in)

### LINKED ISSUE #1072: Footer updates
Simplify the content of the footer & include the FFIEC Home Mortgage Disclosure Act logo with a link to the homepage

wpears:
Updates app to no longer expect to live at `/`

Depends on #1068 
Closes #1071, Closes #1072

Should be reviewed with https://github.com/cfpb/hmda-platform/pull/1618, https://github.com/cfpb/hmda-pub-ui/pull/147, https://github.com/cfpb/hmda-platform-tools/pull/136, https://github.com/cfpb/hmda-platform-auth/pull/229, and https://github.com/cfpb/hmda-homepage set up.


src/common/Header.jsx
```diff
@@ -19,7 +18,24 @@ export const makeNav = (props, page) => {
   let userHeader = (
     <ul className="usa-nav-primary">
       <li>
-        <HomeLink />
+        <a href={window.HMDA_ENV.HOMEPAGE_URL} className="usa-nav-link">
```
awolfe76: I'm not sure we want to include the entire navigation within the filing app.

I'm not completely against it, we had talked about having the app be a sort-of-separate thing altogether, even applying slightly different styling to it (at some point).

Also, within the app as it is today the "Home" link takes you back to the filing app home page (the institutions listing), with these updates it would now take you to the overall home page.

I think it would be more intuitive to someone filing to not have these links to be able to focus on the filing itself, which is what they would want to do once here.

(Keeping the home link of course.)

I think `/filing/`, **when logged out**, would show all the links. But when logged in have them removed. With some other minor changes we can make this pretty clear.

@designlanguage

src/common/Header.jsx
```diff
@@ -65,11 +68,11 @@ const Header = props => {
           <em className="usa-logo-text">
             <Link
               className="usa-nav-link"
-              to={'/'}
+              to={window.HMDA_ENV.APP_SUFFIX}
               title="Home"
               aria-label="Home"
             >
-              <img src="/img/ffiec-logo.png" width="100px" alt="FFIEC" />
+              <img src="/filing/img/ffiec-logo.png" height="27px" alt="FFIEC" />
```
awolfe76: Should we update the content here to be "HMDA Filing Platform" or "HMDA Filing"?

src/common/Footer.jsx
```diff
@@ -14,11 +14,15 @@ const Footer = () => {
               <li className="usa-footer-primary-content">
```
awolfe76, line 14: We also talked about removing a lot of the content from the footer. Again, since someone in the filing app is likely there to do that single task having all of these links available (which are used prior to filing) doesn't seem to help.

We probably should update the other "sites" footers to use this one and update this one to use the other footer.

#### Other footer (eg, pub-ui)

new-footer

src/index.html
```diff
@@ -110,8 +110,8 @@ <h2>Top FAQs</h2>
                     Filers who wish to confirm that their LAR is formatted in
```
awolfe76: I think we can remove the FAQs section completely. This content would better in a different place.

@keileyo @colleenms - can we also remove the [Notice and Consent Banner and PRA statement](https://github.com/cfpb/hmda-platform-ui/pull/1069/files#diff-e249faefed5757034596c5096d33dab6R124) from this page? It will be on the new homepage. Just not sure if one place is enough or if we need it in both.

awolfe76 (changes requested):

src/common/Header.jsx
wpears: Yeah my logic here was that the changed header link on the left still goes to the `/filing/` home along with the `/filing/` link itself and I didn't want to switch the meaning of what "home" meant without also changing the designs.

We did discuss changing the nav look entirely, but I didn't want to do that in this PR to instead focus on getting the hmda-homepage and `/filing/` changes working.

src/common/Header.jsx
wpears: Possibly! I'd be for either.

src/common/Footer.jsx
awolfe76: #1072 will take care of this.

src/common/Header.jsx
awolfe76: Based on the earlier conversation we should remove this change. We won't have all of the navigation within the filing application.

src/common/Header.jsx
awolfe76: We can make this quick update too.

src/index.html
keileyo: PRA & notice and consent need to be on the filing application page. i am down to remove FAQs.

awolfe76 (approved):

PR #1085, opened 2018-10-19
cleaning out authentication

awolfe76:
The start of cleaning out the current authentication for v2. I wanted to start small PRs to walk through this.

Do you see anything, or know of anything, that I'm missing? I haven't deleted any of the files, I've only commented out or removed code at this point.

This will definitely break the build, but it's going to the `v2` branch and the next steps will get it cleaned up and working again.

### Rough next steps

1. Bring in CRA.
    1. Will submit a PR with only dependecy updates
1. Make sure it can build, without authentication, and at least point to the `v2` API using the proxy feature of CRA and relative paths.
1. Clean up styling (`scss -> css`).
1. Final clean up.
1. Bring back auth.

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,22 +1,18 @@
 import React, { Component } from 'react'
 import { connect } from 'react-redux'
-import { getUserManager, signinRedirect } from './utils/redirect.js'
 import ConfirmationModal from './modals/confirmationModal/container.jsx'
-import LoggedOutModal from './modals/loggedOutModal/container.jsx'
-import * as AccessToken from './api/AccessToken.js'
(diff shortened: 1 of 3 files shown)


wpears (approved):
Looks like a good first step. The receipt container will be broken without a user.oidc key being set, but that'll be addressed in one of those cleanup steps :wink:

awolfe76:
Good catch on the receipt container! Figured I would miss something. :)

Just going to go ahead and [point that out here](https://github.com/cfpb/hmda-platform-ui/blob/0103791272c8366c163ef9b03ffc16b13ebaed88/src/submission/ReceiptContainer.jsx#L8).

PR #1086, opened 2018-10-19
Create react app

awolfe76:
### Depends on #1085 

This PR is just to update the dependencies to include CRA and remove all others. Again, this build will be broken but is also to the v2 branch.

There will possibly (likely?) be other dependency updates to get the build process working there but I wanted to keep the other CRA related PRs clean(er).


package.json
```diff
@@ -37,66 +24,20 @@
...
-    "react": "16.2.0",
-    "react-dom": "16.2.0",
+    "react": "16.4.2",
+    "react-dom": "16.4.2",
     "react-dropzone": "4.2.3",
     "react-redux": "5.0.6",
     "react-router": "v3.2.0",
     "react-router-redux": "4.0.7",
     "react-router-scroll": "0.4.4",
+    "react-scripts": "^2.0.5",
```
wpears: Can you pin this?

awolfe76: 👍 777ba83 and 4043bc1

PR #1087, opened 2018-10-19
CRA build

awolfe76:
Depends on #1086 

This gets CRA building and rendering, even though not correctly. Yet. 😉
You can use `yarn start` to see ...

ui

It's missing all styling and still only loading the loading icon (not seen in the screenshot because it's missing the styling too).

More to come.


awolfe76:
There are also a few `warnings` (not `errors`) still when compiling. See #1088.

PR #1092, opened 2018-10-25
mocks api for institutions page

awolfe76:
Gets us the mocked API for now for the institutions page. Needs some CSS cleanup, coming next.

__NOTE:__ After running `yarn start`, go to `http://localhost:3000/institutions`. There needs to be some work done on the routing ... _I think_.

### Screenshot

institutions-bad-style

PR #1093, opened 2018-10-26
Finish styling institutions page

awolfe76:
### Depends on #1092 

Cleans up the submission history, finishing the institutions page.

### Screenshot

fi-complete


wpears (changes requested):
I think you've probably intended this to go against the `v2` branch rather than `master`.

Also, looks like the second `ViewButton` (view completed filing) doesn't keep its white font color on highlight.

Also, the small upload and download links don't have any hover styles.

awolfe76:
Good catch(es)! Should be 👍 now.

PR #1097, opened 2018-10-29
Syntactical and validity

awolfe76:
### Depends on #1096 

As is this PR will render the __suppressed__ edits. 

edit-suppressed

You can see the "normal" edits view by:

1. commenting out the [`return fetch({ suffix: '/edits/${pathObj.edit}' })`](https://github.com/cfpb/hmda-platform-ui/pull/1097/files#diff-46908a2fc227e9faf10d6c95ac926be9R276) and __un__commenting the mocked `getEdit` API response in `api.js, 
1. [in `TableWrapper`, set `suppressEdits` to `false`](https://github.com/cfpb/hmda-platform-ui/pull/1097/files#diff-83651818d873e276120f7b834457f1f8R76).

You will see the loading icon for all the edits except for `S010`. This is ok for now because we just want to be sure the edits are rendering correctly and this does that. Once we're completely hooked up to the back-end the other edits will render correctly.

edits

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
   "dependencies": {
     "detect-browser": "1.7.0",
     "file-saver": "1.3.3",
-    "normalize.css": "^8.0.0",
+    "normalize.css": "8.0.0",
     "react": "16.4.2",
     "react-app-polyfill": "0.1.3",
     "react-dom": "16.4.2",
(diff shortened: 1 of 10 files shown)


wpears (approved):
Works. Opened #1099 & #1100 to be completed in another PR.

PR #1101, opened 2018-10-30
Cra css summary

awolfe76:
Handles the submission page CSS.

submission


src/submission/irs/index.jsx
```diff
@@ -7,6 +7,7 @@ import Alert from '../../common/Alert.jsx'
 import './IRSReport.css'
 
 const renderTotals = props => {
+  console.log(props)
```
wpears: spare log

PR #1102, opened 2018-10-31
Id to lei

awolfe76:
The [API responses](https://github.com/cfpb/hmda-platform/blob/master/docs/v2/api/filing-api.md) changed from using `id` and `institutionId` to using `lei`. This PR takes care of updating the mocked responses and the redux and component code to account for this change.

👀 _But, please double check me_. 👀


wpears:
I'm a little squeamish about referring to the `lei` in the request but then including it in the store as an `id` or `institutionId`, but that's a bigger refactor.

Issue #1104, opened 2018-11-01
Create filing when none exists


awolfe76:
For v2 filings will __not__ exist by default. So if the [`getFiling`](https://github.com/cfpb/hmda-platform-ui/blob/0103791272c8366c163ef9b03ffc16b13ebaed88/src/api/api.js#L18-L20) call returns a `404`, we'll need to `POST` to that endpoint to have it created. This is similar to how we handle submssions.

Currently our test insitution, `{ lei: 'BANK1LEIFORTEST12345', name: 'Bank 1' }`, does not have a filing created and can be used for testing. It's also possible to create more test institutions using the help app if you need to.

See the [updated API docs](https://github.com/cfpb/hmda-platform/blob/df6fa494341cea5e5a7d94db5bac342d3669ae1e/docs/v2/api/filing-api.md#filings) for the `POST` information.

Issue #1083, opened 2018-10-09
API integration


awolfe76:
Integrate new endpoints as they are developed in v2.

The [list of endpoints will be the same as what is listed for v1](https://github.com/cfpb/hmda-platform/blob/bbb6332f07caec8bf0b204a79303269afc4962fc/docs/v1/API.md#institutions-endpoints) with minor updates to the responses, eg `institutionId` or `institution.id` will become `lei`. There may also be a few new endpoints added.

awolfe76:
Closing, all endpoints are 👍. There are a few known things missing/need tweaking, eg the CSV download, but that _shouldn't_ require anything on our end ... if they do a new issue should be created.

PR #1105, opened 2018-11-01
working call for filings using demo proxy

awolfe76:
Makes progress towards #1083 ... hopefully, the rest are this straightforward. 😄 

### Testing

You'll have to add the `proxy` to `package.json` and then run `yarn start`.

### Screenshot

You'll see the loading icon for the second institution, Bank 1. This is because there isn't a filing created for that institution yet. #1104 was created to handle this. That _should_ also remove the error at the top of the page.

---

filngs-404


wpears (approved):
Neat!

PR #1113, opened 2018-11-07
Docker and Helm

awolfe76:
Adds necessary files for v2 deployment.

### DIFF
--- a/kubernetes/hmda-platform-ui/templates/service.yaml
+++ b/kubernetes/hmda-platform-ui/templates/service.yaml
@@ -30,8 +30,8 @@ metadata:
       apiVersion: ambassador/v0
       kind: Mapping
       name: tools_mapping
-      prefix: /tools/
-      rewrite: /tools/
+      prefix: /filing/
+      rewrite: /filing/
       service: {{ include "hmda-platform-ui.fullname" . }}:{{ .Values.service.port }}
 spec:
   type: {{ .Values.ambassador.service.type }}
--- a/kubernetes/hmda-platform-ui/templates/service.yaml
+++ b/kubernetes/hmda-platform-ui/templates/service.yaml
@@ -29,7 +29,7 @@ metadata:
     getambassador.io/config: |
       apiVersion: ambassador/v0
       kind: Mapping
-      name: tools_mapping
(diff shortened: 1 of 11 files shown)


kubernetes/hmda-platform-ui/templates/service.yaml
```diff
@@ -0,0 +1,43 @@
...
+metadata:
+  labels: 
+    service: {{ include "hmda-platform-ui.fullname" . }}-ambassador
+  name: {{ include "hmda-platform-ui.fullname" . }}-ambassador
+  annotations:
+    getambassador.io/config: |
+      apiVersion: ambassador/v0
+      kind: Mapping
+      name: tools_mapping
+      prefix: /tools/
```
wpears: This should probably be `/filing/` or something similar

wpears (changes requested):

kubernetes/hmda-platform-ui/templates/service.yaml
awolfe76: Good catch. I thought I got them all. 😛 

e37b451

kubernetes/hmda-platform-ui/templates/service.yaml
```diff
@@ -0,0 +1,43 @@
...
+kind: Service
+metadata:
+  labels: 
+    service: {{ include "hmda-platform-ui.fullname" . }}-ambassador
+  name: {{ include "hmda-platform-ui.fullname" . }}-ambassador
+  annotations:
+    getambassador.io/config: |
+      apiVersion: ambassador/v0
+      kind: Mapping
+      name: tools_mapping
```
kgudel: `tools_mapping` is already in use. This should be changed to something like `platform_ui_mapping`.

kgudel (changes requested):
Other than one small issue this all builds properly in jenkins and deploys correctly to kubernetes.

kubernetes/hmda-platform-ui/templates/service.yaml
awolfe76: 65ea204

wpears (approved):

PR #1120, opened 2018-11-13
Push Images to Internal Repo and Tag Images on Git Tags

kgudel:
See: https://github.com/cfpb/hmda-pub-ui/pull/210


Jenkinsfile
```diff
@@ -11,15 +11,33 @@ volumes: [
      def gitBranch = repo.GIT_BRANCH
      def shortGitCommit = "${gitCommit[0..10]}"
```
omnipresent07: Unused var `shortGitCommit` can possibly be removed but other than this LGTM 👍

PR #1121, opened 2018-11-14
V2 submission codes

### LINKED ISSUE #1024: Refactor for new status codes
Depends on https://github.com/cfpb/hmda-platform/issues/1457.

After the back-end updates the status codes for v2, we can go through the app can clean up logic because we should be able to base most progress on the status code alone. There shouldn't be a need to check if edits exist to allow for stepping through the app.

awolfe76:
Closes #1024 

This is a bit hard to test everything until we get the full workflow down from the back-end but it looks correct or at least ok for now and we can fix things as they come up. I don't think I missed any places where the code (and checks for edits are used) but have a 👀.


awolfe76:
Also created #1122 to improve messaging here.

src/institutions/Refile.jsx
```diff
@@ -1,14 +1,11 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import RefileButton from '../refileButton/container.jsx'
-import * as STATUS from '../constants/statusCodes.js'
+import { PARSED_WITH_ERRORS, VALIDATING } from '../constants/statusCodes.js'
```
awolfe76, line 4: I've done this in a few places just to make what is actually being used more clear.

src/institutions/ViewButton.jsx
```diff
@@ -1,25 +1,31 @@
...
+  if (code <= CREATED) {
     text = 'Upload your file'
-  } else if (code < STATUS.PARSED_WITH_ERRORS) {
+  } else if (code < PARSED_WITH_ERRORS) {
     text = 'View upload progress'
-  } else if (code === STATUS.PARSED_WITH_ERRORS) {
+  } else if (code === PARSED_WITH_ERRORS) {
     text = 'Review formatting errors'
-  } else if (code < STATUS.SYNTACTICAL_VALIDITY_EDITS) {
+  } else if (code < NO_SYNTACTICAL_VALIDITY_EDITS) {
```
awolfe76, line 24: The `NO_` status happens first but would also tell us that `VALIDATING` has stopped.

src/submission/Nav.jsx
```diff
@@ -26,7 +30,7 @@ export default class EditsNav extends Component {
       upload: {
         isReachable: () => true,
         isErrored: () => this.props.code === PARSED_WITH_ERRORS,
-        isCompleted: () => this.props.code > VALIDATING,
+        isCompleted: () => this.props.code > PARSED_WITH_ERRORS,
```
awolfe76: We should be able to say upload `isCompleted` at `PARSED_WITH_ERRORS`, make sense?

src/submission/Nav.jsx
```diff
@@ -37,8 +41,9 @@ export default class EditsNav extends Component {
           this.props.editsFetched && this.navMap.upload.isCompleted(),
         isErrored: () => this.props.code === SYNTACTICAL_VALIDITY_EDITS,
         isCompleted: () =>
-          this.navMap['syntactical & validity edits'].isReachable() &&
-          this.props.code > SYNTACTICAL_VALIDITY_EDITS,
+          (this.navMap['syntactical & validity edits'].isReachable() &&
+            this.props.code > SYNTACTICAL_VALIDITY_EDITS) ||
+          this.props.code === NO_SYNTACTICAL_VALIDITY_EDITS,
```
awolfe76, line 46: Adding the `NO_` status here (and below) but not sure it's necessary. Seeing it run completely in certain cases will help.

src/submission/signature/index.jsx
```diff
@@ -16,10 +16,7 @@ const showWarning = props => {
 }
 
 const Signature = props => {
-  let isDisabled =
-    props.status.code >= VALIDATED && props.status.code !== SIGNED
-      ? false
-      : true
+  let isDisabled = props.status.code === VALIDATED ? false : true
```
awolfe76: The previous `&&` check seemed redundant. `VALIDATED` is immediately followed by `SIGNED`.

src/submission/upload/DropzoneContent.jsx
```diff
@@ -27,6 +27,8 @@ const DropzoneContent = ({ code, errors, filename, errorFile }) => {
       case STATUS.PARSING:
       case STATUS.PARSED:
       case STATUS.VALIDATING:
+      case STATUS.NO_SYNTACTICAL_VALIDITY_EDITS:
```
awolfe76, line 30: Added these here, but again, not sure if they are necessary.

src/submission/upload/ValidationProgress.jsx
```diff
@@ -1,7 +1,15 @@
 import React, { PureComponent } from 'react'
 import PropTypes from 'prop-types'
 import ProgressText from './ProgressText.jsx'
-import * as STATUS from '../../constants/statusCodes.js'
+import {
+  PARSED_WITH_ERRORS,
+  VALIDATING,
+  UPLOADING
+} from '../../constants/statusCodes.js'
+/* TODO
```
awolfe76, line 9: Just a reminder that we should watch what happens here in certain cases, like a clean file.

src/submission/upload/index.jsx
```diff
@@ -21,12 +25,21 @@ export default class Upload extends Component {
 
   componentDidMount() {
     const { code, pollSubmission } = this.props
+    /* TODO
```
awolfe76, line 28: Just a reminder that we should watch what happens here in certain cases, like a clean file.

src/submission/Nav.jsx
wpears: I don't think that's true, cause they should be stuck on the upload page until validation is complete

src/refileWarning/index.jsx
```diff
@@ -82,23 +85,22 @@ export const getHeading = props => {
...
+  if (props.code >= VALIDATED || props.code < PARSED_WITH_ERRORS) return null
   if (
     props.page === 'syntacticalvalidity' &&
-    !props.syntacticalValidityEditsExist
+    props.code !== SYNTACTICAL_VALIDITY_EDITS
   )
     return null
-  if (props.page === 'quality' && props.qualityVerified) return null
-  if (props.page === 'macro' && props.macroVerified) return null
+  if (props.page === 'quality' && props.code === MACRO_EDITS) return null
```
wpears: this should be `props.code > QUALITY_EDITS`

src/refileWarning/index.jsx
```diff
@@ -82,23 +85,22 @@ export const getHeading = props => {
...
+  if (props.page === 'macro' && props.code === VALIDATED) return null
```
wpears: this should be `props.code > MACRO_EDITS`

src/submission/signature/index.jsx
wpears: I believe it was a safeguard against a user reaching the submission page when the file wasn't ready to be submitted (thus the `>` rather than equality check). The router should prevent this, but this was a failsafe.

wpears (changes requested):
Some inline comments.

Also, if we want to move away from the `syntacticalValidityEditsExist` & `qualityVerified` & `macroVerified` keys in the HOC, we might as well remove them everywhere rather than commenting them out. The alternative would be to put the eg `code === SYNACTICAL_VALIDITY_EDITS` in the HOC rather than the more complicated check, but pairing down passed props is probably a good thing since the checks are now straightforward.

awolfe76:
56836e2 should take care of the comments.

wpears (approved):

PR #1123, opened 2018-11-16
API response updates AND update to allow fetching of new filings sooner

awolfe76:
See inline comments

### DIFF
--- a/src/actions/fetchCurrentFiling.js
+++ b/src/actions/fetchCurrentFiling.js
@@ -1,11 +1,15 @@
 import fetchFiling from './fetchFiling.js'
+import fetchNewFiling from './fetchNewFiling.js'
 
-export default function fetchCurrentFiling(filings) {
+export default function fetchCurrentFiling(institution) {
   return (dispatch, getState) => {
     const period = getState().app.filingPeriod
-    const filing = filings.filter(filing => {
+    const filing = institution.filings.filter(filing => {
       return filing.period === period
     })[0]
-    return dispatch(fetchFiling(filing))
+    if (filing) return dispatch(fetchFiling(filing))
+    return dispatch(
+      fetchNewFiling({ lei: institution.institution.lei, period: period })
+    )
   }
 }
--- a/src/actions/fetchCurrentFiling.js
+++ b/src/actions/fetchCurrentFiling.js
@@ -1,11 +1,21 @@
 import fetchFiling from './fetchFiling.js'
+import fetchNewFiling from './fetchNewFiling.js'
 
-export default function fetchCurrentFiling(filings) {
+export default function fetchCurrentFiling(institution) {
   return (dispatch, getState) => {
     const period = getState().app.filingPeriod
-    const filing = filings.filter(filing => {
+
+    const filing = institution.filings.filter(filing => {
       return filing.period === period
     })[0]
-    return dispatch(fetchFiling(filing))
+
+    if (filing) return dispatch(fetchFiling(filing))
+
+    return dispatch(
+      fetchNewFiling({
+        lei: institution.institution.lei,
+        period: period
+      })
+    )
   }
 }
--- a/src/actions/fetchCurrentFiling.js
+++ b/src/actions/fetchCurrentFiling.js
@@ -1,15 +1,11 @@
 import fetchFiling from './fetchFiling.js'
-import fetchNewFiling from './fetchNewFiling.js'
 
-export default function fetchCurrentFiling(institution) {
+export default function fetchCurrentFiling(filings) {
   return (dispatch, getState) => {
     const period = getState().app.filingPeriod
-    const filing = institution.filings.filter(filing => {
+    const filing = filings.filter(filing => {
       return filing.period === period
     })[0]
-    if (filing) return dispatch(fetchFiling(filing))
-    return dispatch(
-      fetchNewFiling({ lei: institution.institution.lei, period: period })
-    )
+    return dispatch(fetchFiling(filing))
   }
 }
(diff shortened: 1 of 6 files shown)


src/actions/fetchInstitution.js
```diff
@@ -17,8 +18,15 @@ export default function fetchInstitution(institution, fetchFilings = true) {
             throw new Error(json && `${json.status}: ${json.statusText}`)
           }
           dispatch(receiveInstitution(json))
-          if (json.filings && fetchFilings) {
+          if (json.filings.length !== 0 && fetchFilings) {
```
awolfe76, line 23: The response here returns `filings: []` so we can see the `length !== 0` and fetch the current filing OR create the new one with the current `lei` and `filingPeriod`.

src/api/api.js
```diff
@@ -5,67 +5,15 @@ export function getInstitutions() {
   return new Promise(resolve => {
     resolve({
       institutions: [
-        { lei: 'B90YWS6AFX2LGWOXJ1LD', name: 'Bank 0' },
-        { lei: 'BANK1LEIFORTEST12345', name: 'Bank 1' }
+        { lei: '01KWVG908KE7RKPTNP46', name: 'Bank 0' },
```
awolfe76, line 8: You can swap these for a new institution to test out the logic locally. Let me know if you need another `lei` to test.

src/api/api.js
```diff
@@ -5,67 +5,15 @@ export function getInstitutions() {
...
-            status: { code: 1, message: 'not-started' },
-            end: 0,
-            start: 0,
-            period: '2017'
-          }
-        ]
-      })
-    })
-  }
+  return fetch({ pathname: `/institutions/${id}` })
```
awolfe76, line 16: Just cleaning out the mocked response.

src/reducers/institutions.js
```diff
@@ -41,7 +41,7 @@ export default (state = defaultInstitutions, action) => {
           ...state.institutions,
           [action.institution.lei]: {
             isFetching: false,
-            name: action.institution.name,
+            name: action.institution.respondent.name,
```
awolfe76, line 44: The response here is now:

```json
"respondent":{"name":"name","state":"ST","city":"CITY"}
```

src/submission/parseErrors/index.jsx
```diff
@@ -66,7 +66,7 @@ export const renderLarErrors = ({ larErrors, ...props }) => {
           return larErrorObj.errorMessages.map((message, i) => {
             return (
               <tr key={i}>
-                <td>{larErrorObj.lineNumber}</td>
+                <td>{larErrorObj.rowNumber}</td>
```
awolfe76, line 69: API response changed here too.

src/submission/parseErrors/index.jsx
```diff
@@ -86,12 +86,10 @@ class ParseErrors extends Component {
 
   render() {
     const props = this.props
+
     if (!props.fetched) return <Loading />
 
-    const total =
```
awolfe76, line 91: We have this total already ... plus I was seeing it as off by minus 1. Let me know if you see this differently. This seems to get it right.

src/submission/parseErrors/index.jsx
```diff
@@ -103,7 +101,7 @@ class ParseErrors extends Component {
         <header>
           {!props.pagination ? null : (
             <h2>
-              {total} {errorText} with Formatting Errors
+              {props.pagination.total} {errorText} with Formatting Errors
```
awolfe76, line 104: From above, we can just use the total here.

src/actions/fetchCurrentFiling.js
```diff
@@ -1,11 +1,21 @@
 import fetchFiling from './fetchFiling.js'
+import fetchNewFiling from './fetchNewFiling.js'
 
-export default function fetchCurrentFiling(filings) {
+export default function fetchCurrentFiling(institution) {
```
awolfe76, line 4: Accepts the whole `institution` so [we have the `lei`](https://github.com/cfpb/hmda-platform-ui/pull/1123/files#diff-845f5e1adaa96e28a8010b49daa64eb6R16) to make a call for a new filing if necessary.

src/actions/fetchCurrentFiling.js
```diff
@@ -1,11 +1,21 @@
...
   return (dispatch, getState) => {
     const period = getState().app.filingPeriod
-    const filing = filings.filter(filing => {
+
+    const filing = institution.filings.filter(filing => {
       return filing.period === period
     })[0]
-    return dispatch(fetchFiling(filing))
+
+    if (filing) return dispatch(fetchFiling(filing))
```
awolfe76, line 12: This catches the case where we would have `filings` but none that match the current `period`, eg a filing for 2016 or 2017.

In that case, we still need to [create a new filing for the current `period`](https://github.com/cfpb/hmda-platform-ui/pull/1123/commits/a5441bc45acf304d763474e1e05d1d4067293667#diff-845f5e1adaa96e28a8010b49daa64eb6R14).

src/institutions/index.jsx
```diff
@@ -46,7 +46,7 @@ const _whatToRender = ({ filings, filingPeriod, institutions, submission }) => {
     const institution = institutions.institutions[key]
     const institutionFilings = filings[institution.id]
 
-    if (!institutionFilings) {
+    if (!institutionFilings || !institutionFilings.filing) {
```
awolfe76, line 49: If the call to create a new filing fails, this will let us render the UI a little cleaner (rather than just the loading icon).

create-failed

PR #1128, opened 2018-11-20
make some images publicly available so keycloak has access to them

### LINKED ISSUE #1127: Add Images to Public Folder
Keycloak needs to be able to grab some images from the `hmda-platform-ui` application:
 - `favicon-57.png`
 - `ffiec-logo.svg`

awolfe76:
Closes #1127 

The `src/images` are also _technically_ available, but they include a hash that updates on each build. Putting these images in public allows for a consistent path.


wpears:
We should probably include these images in https://github.com/cfpb/hmda-platform/tree/master/kubernetes/keycloak/theme-provider/themes/hmda/login/resources/img rather than hotlinking to them from keycloak, so that it can stand on its own if the structure of the ui changes

PR #1142, opened 2018-11-23
get institutions based on token data

### LINKED ISSUE #1134: Get available institutions
Now that #1132 is merged, we can get the list of institutions (`lei`s) to make the correct API calls for the institution and filing data.

_This should finish removing the [mocked API responses](https://github.com/cfpb/hmda-platform-ui/blob/b48869078803c224b1ea519ff6ac5b89f95e6110/src/api/api.js#L3-L13)._

awolfe76:
Closes #1134 

See inline comments


src/actions/fetchInstitutions.js
```diff
@@ -3,7 +3,7 @@ import receiveInstitutions from './receiveInstitutions.js'
 import receiveError from './receiveError.js'
 import hasHttpError from './hasHttpError.js'
 import requestInstitutions from './requestInstitutions.js'
-import { getInstitutions } from '../api/api.js'
+//import { getInstitutions } from '../api/api.js'
```
awolfe76: This action is no longer used at all. We can remove it but wanted to let it for now. Let me know what you think.

src/api/api.js
```diff
@@ -1,17 +1,5 @@
 import { fetch } from './fetch.js'
 
-export function getInstitutions() {
```
awolfe76, line 3: We no longer make a call for the list of institutions. We use the token data instead.

src/institutions/container.jsx
```diff
@@ -1,12 +1,26 @@
...
+import fetchEachInstitution from '../actions/fetchEachInstitution.js'
+import receiveInstitutions from '../actions/receiveInstitutions.js'
 import Institutions from './index.jsx'
+import { getKeycloak } from '../utils/keycloak.js'
 
 export class InstitutionContainer extends Component {
   componentDidMount() {
     if (!this.props.institutions.fetched && !this.props.institutions.isFetching)
-      this.props.dispatch(fetchInstitutions())
+      this.props.dispatch(requestInstitutions())
```
awolfe76, line 12: Left this in because it basically just triggers the loading icon.

src/institutions/container.jsx
```diff
@@ -1,12 +1,26 @@
...
+    const leis = getKeycloak().tokenParsed.lei.split(',')
+
+    // create the expected objects from the array, institutions = [{lei: lie}, {lei: lei}]
+    let institutions = []
+    leis.forEach(lei => {
+      institutions.push({ lei: lei })
+    })
+    this.props.dispatch(fetchEachInstitution(institutions))
+
+    this.props.dispatch(receiveInstitutions())
```
awolfe76, line 24: Same as above ... left in because it removes the loading icon.

src/institutions/container.jsx
```diff
@@ -1,12 +1,26 @@
...
 import Institutions from './index.jsx'
+import { getKeycloak } from '../utils/keycloak.js'
 
 export class InstitutionContainer extends Component {
   componentDidMount() {
     if (!this.props.institutions.fetched && !this.props.institutions.isFetching)
-      this.props.dispatch(fetchInstitutions())
+      this.props.dispatch(requestInstitutions())
+
+    const leis = getKeycloak().tokenParsed.lei.split(',')
```
awolfe76: This, and the few lines below, are now how we fetch the institutions ...
- get the `lei`s from the token
- build the array of objects
- call `fetchEachInstitution` with that array

src/institutions/index.jsx
```diff
@@ -46,10 +46,7 @@ const _whatToRender = ({ filings, filingPeriod, institutions, submission }) => {
     const institution = institutions.institutions[key]
     const institutionFilings = filings[institution.id]
 
-    if (!institutionFilings || !institutionFilings.filing) {
-      // there are no filings
-      return <Institution key={i} institution={institution} />
-    } else if (!institutionFilings.fetched) {
+    if (!institutionFilings || !institutionFilings.fetched) {
```
awolfe76, line 53: Doing this prevents a quick rendering of the "problem initializing your filing" error.

PR #1156, opened 2018-11-28
User routing

### LINKED ISSUE #1144: Forward logged in users to institutions page

wpears:
Closes #1144

### DIFF
--- a/src/common/Header.jsx
+++ b/src/common/Header.jsx
@@ -19,7 +19,7 @@ export const logOutHandler = e => {
 
 export const getLink = () => {
   if(getKeycloak().authenticated) return '/filing/institutions'
-  return '/filing'
+  return '/filing/'
 }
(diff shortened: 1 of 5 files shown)


src/common/Header.jsx
```diff
@@ -17,11 +17,16 @@ export const logOutHandler = e => {
   logout()
 }
 
+export const getLink = () => {
+  if(getKeycloak().authenticated) return '/filing/institutions'
+  return '/filing'
```
awolfe76: We should add the trailing slash to this `return`.

```suggestion
  return '/filing/'
```

_It's not required for other paths, just the root._

awolfe76 (changes requested):
Just the one thing with the trailing slash.

awolfe76 (approved):

PR #1159, opened 2018-11-28
Provide the correct key when requesting institution for header name

### LINKED ISSUE #1158: Call for "undefined" institution
When #1157 happens, if you refresh the browser you'll see a call to `http://localhost:3000/v2/filing/institutions/undefined`, which of course responds with a `404`.

Unclear at the moment why the `lei` (`id`) is `undefined`, need to debug.

This call is happening [submission container](https://github.com/cfpb/hmda-platform-ui/blob/0103791272c8366c163ef9b03ffc16b13ebaed88/src/submission/container.jsx#L74-L81).

wpears:
Closes #1158

### DIFF
--- a/src/submission/container.jsx
+++ b/src/submission/container.jsx
@@ -103,7 +103,7 @@ class SubmissionContainer extends Component {
     const id = this.props.params.institution
 
     if (!this.props.institutions.institutions[id]) {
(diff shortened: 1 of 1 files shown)


awolfe76 (approved):
😄

PR #1173, opened 2018-11-29
Edit details

awolfe76:
This will work as the API is right now, but https://github.com/cfpb/hmda-platform/pull/2265 will update the response. __We should hold off until that is merged and make any other changes needed.__


wpears (approved):
Works as intended under current API response

PR #1184, opened 2018-12-04
WIP - IE styling

awolfe76:
# WIP

@wpears - this is a very quick proposal to make fix the width issue in IE 11. See the inline comment.

I've left it as "WIP" because if this is ok we can do something similar to the other "pages".

### DIFF
--- a/src/institutions/Institutions.css
+++ b/src/institutions/Institutions.css
@@ -1,10 +1,15 @@
 .Institutions {
-  max-width: 800px;
+  margin-top: 3em;
+}
+
+.Institutions > * {
   margin: 0 auto;
+  max-width: 800px;
 }
 
 .Institutions .institution {
   margin-top: 2em;
+  max-width: 850px;
 }
 
 .Institutions .institution .current-status {
@@ -57,6 +62,10 @@
 
 .Institutions p.multi-message {
   margin-top: 2.5em;
+}
+
+.Institutions .alert-info {
+  margin-top: 2em;
 }
 
 .floatingIcon + p.multi-message {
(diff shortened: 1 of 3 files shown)


src/institutions/Institutions.css
```diff
@@ -1,10 +1,15 @@
 .Institutions {
-  max-width: 800px;
+  margin-top: 3em;
+}
+
+.Institutions > * {
```
awolfe76, line 5: Here we just add the `max-width` to all the children directly.

wpears (changes requested):
This doesn't fix any of the submissions pages, right? Seems to me like we just need to pull in the grid css that got dropped when uswds was pulled out that is already tuned to work in ie11.

If not, the `.Institutions > *` workaround needs to apply to the submissions pages as well

awolfe76:
Yeah, that's exactly right. That's why 
> I've left it as "WIP" because if this is ok we can do something similar to the other "pages".

😄 

Just wanted to see what you thought to start. I'll get the rest of them now.

wpears:
Aha! I thought you were referring to other "pages" meaning tools, pub-ui, homepage, and hmda-help, which I believe will need similar grid adjustments

wpears:
Oh, also the auth templates in hmda-platform will need associated fixes (header, mainly)

src/institutions/Institutions.css
wpears: fwiw, probably the quicker solution is to just add the relevant grid styling, since we're still including the `usa-grid` class

awolfe76: I'm not sure that's the case throughout the app, but that's what I'll try to do tomorrow first thing.

PR #1189, opened 2018-12-05
updates to have the root of the v2 filing UI be /filing/2018

### LINKED ISSUE #1180: Serve from a different root
We don't need to start this yet, need to write up some more details, but this is likely to happen.

Due to the 2017 filing app UI being served from `/filing` we'll need to make the 2018 UI (`v2`) be served from `/2018/filing`.

This will require us to update the [`homepage` in `package.json`](https://github.com/cfpb/hmda-platform-ui/blob/25f770a916f6bb781ec8ba2b2f2a173ff522d289/package.json#L55) to be `/2018/filing`.

It will also require us to update any links that point to that location, eg [the header navigation](https://github.com/cfpb/hmda-platform-ui/blob/25f770a916f6bb781ec8ba2b2f2a173ff522d289/src/common/Header.jsx#L20-L22).

_This will likely cause updates the other UI repos._

awolfe76:
Closes #1180 

This will serve the v2 UI from `/filing/2018`. This is to allow us to run both v1 (2017) and v2 (2018) together. We will be routing every all `/filing/2018` to v2 and all `/filing/` to v1.

This also includes updates to the `Dockerfile` and helm chart.

### DIFF
--- a/kubernetes/hmda-platform-ui/templates/service.yaml
+++ b/kubernetes/hmda-platform-ui/templates/service.yaml
@@ -30,8 +30,8 @@ metadata:
       apiVersion: ambassador/v0
       kind: Mapping
       name: platform_ui_mapping
-      prefix: /filing/
-      rewrite: /filing/
+      prefix: /filing/2018/
+      rewrite: /filing/2018/
       service: {{ include "hmda-platform-ui.fullname" . }}:{{ .Values.service.port }}
 spec:
   type: {{ .Values.ambassador.service.type }}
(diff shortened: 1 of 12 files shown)


kubernetes/hmda-platform-ui/templates/service.yaml
```diff
@@ -30,8 +30,8 @@ metadata:
       apiVersion: ambassador/v0
       kind: Mapping
       name: platform_ui_mapping
-      prefix: /filing/
-      rewrite: /filing/
+      prefix: /filing/2018/
```
wpears: Will the trailing slash here cause issues?

wpears (approved):

PR #1216, opened 2018-12-19
Switch status code handling to new backend format

### LINKED ISSUE #1215: Update status code logic to match backend

wpears:
Closes #1215

### DIFF
--- a/src/actions/pollForProgress.js
+++ b/src/actions/pollForProgress.js
@@ -6,7 +6,6 @@ import { getLatestSubmission } from '../api/api.js'
 import { error } from '../utils/log.js'
 import {
   PARSED_WITH_ERRORS,
-  VALIDATING,
   SYNTACTICAL_VALIDITY_EDITS,
   NO_MACRO_EDITS,
   MACRO_EDITS,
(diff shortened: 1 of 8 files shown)


src/actions/pollForProgress.js
```diff
@@ -8,7 +8,7 @@ import {
   PARSED_WITH_ERRORS,
   VALIDATING,
```
awolfe76: I think we can remove `VALIDATING` here.

src/submission/NavButton.jsx
```diff
@@ -12,7 +12,7 @@ import {
```
awolfe76, line 11: I think we can remove `QUALITY_EDITS` ☝️.

awolfe76 (changes requested):
Looks good, just 2 small `warnings` I noticed from CRA.

Waiting to merge until we figure out a few things on the back-end to avoid more confusion.

awolfe76:
@wpears - I just noticed that the "retrieving edits" message looks like it's showing up too early too. We're still making calls to `/latest` and that message is there. Probably related to updates in this PR. IF you try a larger file you should see it.

awolfe76 (approved):

PR #1233, opened 2019-01-03
Update token

### LINKED ISSUE #1231: Entity Stream Truncation
This is related to https://github.com/cfpb/hmda-platform/issues/2410

When uploading a large file (1M row file) from the UI, the request seems to get cancelled in the middle and backend receives error `Entity Stream Truncation`. If we upload a file directly to the backend API (bypassing UI), then we do not receive `Entity Stream Truncation` error. 

This might have something to do with the Nginx timeouts.

awolfe76:
Closes #1231 

There weren't ambassador errors because the call to update the token wasn't really failing, it just wasn't updating the token because the token hadn't expired ... or wasn't close enough to expiring.

This update

- uses the `error` to handle an actual error and only calls `login()` if the update errors,
- sets the [`updateToken(minValidity)`](https://www.keycloak.org/docs/3.0/securing_apps/topics/oidc/javascript-adapter.html#_updatetoken_minvalidity)(look for `minValidity`) to 60 seconds which means that if the token is due to expire within 60 seconds it will be refreshed,
- and makes the call to update the token a little more frequently to help ensure it doesn't expire.

### DIFF
--- a/src/utils/keycloak.js
+++ b/src/utils/keycloak.js
@@ -25,19 +25,23 @@ const login = (path = '/filing/2018/institutions') => {
 
 const refresh = () => {
   const updateKeycloak = () => {
+    console.log('updateKeycloak')
+    console.log(+(keycloak.tokenParsed.exp + '000') - Date.now() - 10000)
     setTimeout(() => {
       keycloak
-        .updateToken(60)
+        .updateToken(20)
         .success(success => {
           if (success) {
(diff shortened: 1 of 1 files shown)


awolfe76:
This PR also helps toward https://github.com/cfpb/hmda-platform/issues/2410.

wpears (approved):
🥇

PR #1243, opened 2019-01-15
add platform unavailable message

awolfe76:
Quickly add an "unavailable" message to the institution landing page. This removes all available actions a filer could take and simply shows the `<Alert />`.


keileyo:
message looks good.

PR #1253, opened 2019-06-11
Remove Failing Shadow Dependency and Upgrade Nginx to 1.16

kgudel:


kgudel:
Closed in favor of a different approach

PR #1256, opened 2019-06-25
Update service.yaml

zencircle:
Remove nodeport breaks current production configuration


zencircle:
- adds ambassador_id to the mapping configuration
- removes nodeports from ambassador service
- update application service from nodeport to ClusterIP

PR #1260, opened 2019-07-29
Update new api end points

### LINKED ISSUE #1254: Update platform-ui for multiple years
Now that the backend APIs are changed for multiple years, the front end can use these APIs to submit the files for either 2018 or 2019. 

- [x] Change the year on the main page
- [x] Change `/institutions/<lei>/` to `/institutions/<lei>/year/<year>`
- [x] Following the `year` for all API Calls

BarakStout:
Closes #1254

### DIFF
--- a/src/common/Footer.jsx
+++ b/src/common/Footer.jsx
@@ -5,8 +5,10 @@ import logo from '../images/ffiec-logo.svg'
 import { getKeycloak } from '../utils/keycloak.js'
 
 export const getLink = () => {
-  if (getKeycloak().authenticated) return '/filing/2018/institutions'
-  return '/filing/2018/'
+  var year = window.location.pathname.substring(8,12)
+  if (getKeycloak().authenticated) return '/filing/'+year+'/institutions'
+  return '/filing/'+year+'/'
+
 }
 
 const Footer = () => {
--- a/src/common/Header.jsx
+++ b/src/common/Header.jsx
@@ -17,16 +17,17 @@ export const logOutHandler = e => {
   logout()
 }
 
-export const getLink = () => {
-  if (getKeycloak().authenticated) return '/filing/2018/institutions'
-  return '/filing/2018/'
+export const getLink = (path) => {
+  let year = path.substring(8,12)
+  if (getKeycloak().authenticated) return '/filing/'+year+'/institutions'
+  return '/filing/'+year+'/'
 }
 
 export const makeNav = (props, page) => {
   let userHeader = (
     <ul className="nav-primary">
       <li>
-        <Link to={getLink()} className="nav-link">
+        <Link to={getLink(props['pathname'])} className="nav-link">
           Filing Home
         </Link>
       </li>
@@ -57,7 +58,7 @@ const Header = props => {
           <span className="logo-text">
             <Link
               className="nav-link"
-              to={getLink()}
+              to={getLink(props['pathname'])}
               title="Home"
               aria-label="Home"
             >
--- a/src/home/index.jsx
+++ b/src/home/index.jsx
@@ -4,17 +4,19 @@ import { login, register } from '../utils/keycloak.js'
 import './Home.css'
 
 const Home = () => {
+  let year = window.location.pathname.substring(8,12)
   return (
     <main className="Home" id="main-content">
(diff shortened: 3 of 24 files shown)


src/institutions/index.jsx
```diff
@@ -79,7 +82,7 @@ const _whatToRender = ({ filings, filingPeriod, institutions, submission }) => {
 export default class Institutions extends Component {
   render() {
     const { error, filingPeriod } = this.props
-
+    console.log(this.props)
```
wpears, line 85: Stray log

src/institutions/container.jsx
```diff
@@ -17,8 +17,10 @@ export class InstitutionContainer extends Component {
     // create the expected objects from the array, institutions = [{lei: lie}, {lei: lei}]
     let institutions = []
     leis.forEach(lei => {
-      institutions.push({ lei: lei })
+      institutions.push({ lei: lei, year: 2018 })
+      institutions.push({ lei: lei, year: 2019 })
```
wpears, line 21: I don't think this is appropriate here, we should use the values returned in the keycloak object to associate leis with institutions rather than create these associations here.

src/common/Footer.jsx
```diff
@@ -5,8 +5,10 @@ import logo from '../images/ffiec-logo.svg'
 import { getKeycloak } from '../utils/keycloak.js'
 
 export const getLink = () => {
-  if (getKeycloak().authenticated) return '/filing/2018/institutions'
-  return '/filing/2018/'
+  var year = window.location.pathname.substring(8,12)
```
wpears, line 8: I think instead of grabbing the year out of the pathname via `substring`, we should just pass the the `year` down to the component or access it from the redux store.

src/common/Header.jsx
```diff
@@ -17,16 +17,17 @@ export const logOutHandler = e => {
   logout()
 }
 
-export const getLink = () => {
-  if (getKeycloak().authenticated) return '/filing/2018/institutions'
-  return '/filing/2018/'
+export const getLink = (path) => {
+  let year = path.substring(8,12)
```
wpears, line 21: see substring comment

src/home/index.jsx
```diff
@@ -4,17 +4,19 @@ import { login, register } from '../utils/keycloak.js'
 import './Home.css'
 
 const Home = () => {
+  let year = window.location.pathname.substring(8,12)
```
wpears, line 7: see substring comment

src/submission/router.jsx
```diff
@@ -71,14 +71,15 @@ export class SubmissionRouter extends Component {
...
+    const { lei, filing, year } = this.props.params
     return browserHistory.replace(
-      `/filing/2018/${lei}/${filing}/${splat}`
+      `/filing/${year}/${lei}/${filing}/${splat}`
     )
   }
 
   goToAppHome() {
-    return browserHistory.replace('/filing/2018/')
+    var year = window.location.pathname.substring(8,12)
```
wpears, line 81: see substring comment

src/reducers/filingPeriod.js
```diff
@@ -3,6 +3,7 @@ import { UPDATE_FILING_PERIOD } from '../constants'
  * Set the default current filing period
  */
 export default (state = '2018', action) => {
+  state = window.location.pathname.substring(8,12)
```
wpears, line 6: see substring comment

src/utils/keycloak.js
```diff
@@ -53,8 +53,9 @@ const register = () => {
 }
 
 const logout = () => {
+  var year = window.location.pathname.substring(8,12)
```
wpears, line 56: see substring comment

wpears (changes requested):
Outside of the things noted inline, I think we'd be greatly served by combining `year` and `filingPeriod` route vars. If we're versioning the platform-ui per filing period or providing the year/filing period as the first argument to filter on, then it makes sense to just have one of these in the url and only one reflected in the store.

Otherwise, right now we're essentially tracking two variables that are identical and using them at different times which could lead to strange inconsistencies depending on how we're updating these variables as a user switches years.

wpears:
Continued in #1264

PR #1264, opened 2019-08-07
Multi year support/refactor

wpears:
Continuing @BarakStout's work in #1260, this completes the transition to multiyear support (undoing all the gross 2018 hardcoding and using the existing `filingPeriod` throughout the app for consistency).


BarakStout (approved):
LGTM 👍

PR #1270, opened 2019-08-15
K8s multi

wpears:
This is an alternative implementation of https://github.com/cfpb/hmda-platform-ui/pull/1269:

Things different in this approach:
  - from a kubernetes perspective, it just forwards anything with a /filing prefix to the filing app and ignores the rest of the path
 - the create-react-app homepage and nginx file locations are set to /filing/build. This means that static file requests will come to `.../filing/build/...` (I would have preferred this to be just `/filing` but didn't want to deal with the routing rules for the 2017 platform)
 - the url structure remains the same as in 2018, meaning that people's bookmarks will still be valid without introducing/maintaining any other redirects.
 - keeps the year selector on the homepage, which I think is useful as outlined [here](https://github.com/cfpb/hmda-platform/issues/2864#issuecomment-519676649)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,7 @@ FROM nginx:1.16-alpine
 ENV NGINX_USER=svc_nginx_hmda
 RUN rm -rf /etc/nginx/conf.d
 COPY nginx /etc/nginx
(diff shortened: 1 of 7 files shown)


BarakStout (commented):
LGTM

PR #1277, opened 2019-09-25
Switch edit table header from loan/application number to ULI

### LINKED ISSUE #1275: Change Loan/Application Number label to ULI during edits stage
Currently, during the edits stageof the platform, Loan/Application Number is the label of the column displaying the ULI. Users have had trouble interpreting what this means as Loan/Application Number is not the same as ULI to them. The column label should be changed to ULI.

wpears:
Closes #1275


BarakStout (approved):
LGTM 👍

PR #1292, opened 2019-10-03
Prevent doubled institutions list from being build and passed to fetchEachInstitution

### LINKED ISSUE #1289: Extra/redundant POST calls being done from UI
It appears that the UI is making redundant POST calls to the filing endpoint which is causing 400 errors.

wpears:
Closes #1289


BarakStout (approved):
LGTM 👍

PR #1294, opened 2019-10-16
Display all 0 values in the AUS tables as the phrase (blank)

### LINKED ISSUE #1286: Edits: AUS and AUS result values should be displayed as (Blank)
When there's nothing entered in the field, currently 0's are displayed. Instead of 0's, the message (Blank) should be displayed. 

image

PatrickGoRaft:
closes #1286

### DIFF
--- a/src/submission/edits/TableCell.jsx
+++ b/src/submission/edits/TableCell.jsx
@@ -3,10 +3,9 @@ import PropTypes from 'prop-types'
 
 const EditsTableCell = props => {
   let cellContent = props.cell
-  if (props.cell === '') cellContent = <em>(blank)</em>
+  if (props.cell === ''|| props.cell === '0') cellContent = <em>(blank)</em>
   return <td>{cellContent}</td>
 }
-
 EditsTableCell.propTypes = {
   cell: PropTypes.oneOfType([
     PropTypes.string,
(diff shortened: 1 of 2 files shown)


wpears:
This doesn't actually solve the ticket, as the proposed change will change **ALL** `0`s to `(blank)` when the intent is to just do it for the AUS tables.

Personally I think anything involving the clarity of the content of the API response should be fixed at the API level since some institutions will want to file using just the API.

If that is unrealistic, the change on the frontend would likely consist of creating a list of AUS edits in an object (ie `{V696-1: 1,...}`) and checking presence in that list before returning `(blank)` when encountering a 0.

This will either have to be done at a higher level than `TableCell` (which doesn't know what the edit is) or the edit name will need to be passed down to `TableRow` and `TableCell`.

wpears (changes requested):
(see above comment)

PatrickGoRaft:
You are correct my good sir, I was just trying some thing's out between changing the front end and backend. I was going through the component hierarchy to implement what youre suggesting rather than tweaking the results of the API call. 

I was going to add a flag in the event a table header contained a column with a name AUS and pass this flag down to the cell component.

PatrickGoRaft:
@wpears  I've implemented the changes we discussed ,since  `TableRow.jsx` contains the column name data via field.name I'm just passing that down to `TableCell.jsx`

wpears (approved):

PR #1319, opened 2019-11-12
Fix: IE11 doesn't support Object.entries

### LINKED ISSUE #1314: Users are seeing blank screen upon selecting "Review Formatting Errors" in IE
Upon users selecting "Review Formatting Errors" on the Filing screen, 
image

Filers that use Internet Explorer are seeing a blank screen.
image
 I have confirmed that this issue is happening only in IE.

meissadia:
Closes #1314

### DIFF
--- a/src/submission/parseErrors/index.jsx
+++ b/src/submission/parseErrors/index.jsx
@@ -10,13 +10,12 @@ import './ParseErrors.css'
 const RefileWarning = submissionProgressHOC(RefileWarningComponent)
 
 export const renderTSErrors = ({ transmittalSheetErrors }) => {
-
   if (transmittalSheetErrors.length === 0) return null
-  let uliNeeded=false
-  let tsErrorMessages =errorResponseParser(transmittalSheetErrors,uliNeeded)
+  let uliNeeded = false
+  let tsErrorMessages = errorResponseParser(transmittalSheetErrors, uliNeeded)
 
   return (
-    <table width="100%">
+    <table width='100%'>
       <caption>
         <h3>Transmittal Sheet Errors</h3>
         <p>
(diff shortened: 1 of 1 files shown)


wpears:
Hehe, the source of the problem snuck in with the formatting error page revamp, I remember thinking the `for...of` with the `.entries()` method called on an array was a bit odd since it was being called inside a forEach, so it seemed like a bit of a stylistic mismatch... but I was already being pretty nitpicky in the review so I didn't flag it, woops :)

I think my preferred solution would be to just update code that's calling `entries` to use `forEach` like I've done in [this branch](https://github.com/cfpb/hmda-platform-ui/blob/remove-object-entries/src/submission/parseErrors/index.jsx#L131-L149).

If you're a big `.entries()` fan though and want to use it in the future I'd be open to polyfilling it, though I think there might be a more handsoff approach via `babel-polyfill`.

meissadia:
Wish I'd seen your comment much earlier!  I had identified the .entries() call as the issue but didn't correctly translate it when I converted to a .forEach() which led me down a very long path of googling and unnecessary hoop-jumping...

Thanks for the help @wpears

src/submission/parseErrors/index.jsx
```diff
@@ -102,51 +97,61 @@ class ParseErrors extends Component {
...
-      );
-    }
-})
-return apiErrorMessages
+  let apiErrorMessages = []
+
+  errorResponse.forEach(function(errorsFound, i) {
+    if (errorsFound && errorsFound.errorMessages)
+      errorResponse.forEach(function(errorsFound, i) {
+        if (errorsFound && errorsFound.errorMessages) {
```
wpears: Looks like these two lines were doubled, there should only be one set of

```
errorResponse.forEach(function(errorsFound, i) {
    if (errorsFound && errorsFound.errorMessages) {
```

wpears (changes requested):

src/submission/parseErrors/index.jsx
meissadia: 🤕

wpears (approved):

PR #1325, opened 2019-11-21
Fix Institution progress, Nav, router, and submission button

wpears:
… when status code is MACRO_EDITS when macro edits don't actually exist.

This is a patch for a backend issue, but the backend issue is likely being fixed, making this change unnecessary


wpears:
This is, indeed, being fixed on the backend

